PR #71, opened 2014-05-26
Adding safe-snapshot-for-non-boot-disk

palladius:
This script allows one user from within a VM to take a Snapshot for non-root disk.


zmarano:
Hey Riccardo, Is this PR still relevant?

PR #145, opened 2015-02-10
Fix set-hostname under Debian Jessie.

malcolmr:
This script is intended to be used as a dhclient exit hook, sourced (not executed) by `/sbin/dhclient-script`.

The version of `dhclient-script` included in Debian Wheezy uses `/bin/bash` as the shell to run these rather than, as documented, `/bin/sh` (that is, dash).  This is fixed in Jessie, but that fix breaks this script, which depends upon bash.

This change fixes the script to work under dash and bash equally, and also removes the executable bit (as the script is intended to be sourced, not executed directly, and having an executable script with a shebang line just confuses the issue).

### DIFF
--- a/google-startup-scripts/usr/share/google/set-hostname
+++ b/google-startup-scripts/usr/share/google/set-hostname
@@ -1,4 +1,3 @@
-#! /bin/bash
 # Copyright 2013 Google Inc. All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -15,7 +14,7 @@
 
 # Deal with a new hostname assignment.
 
-if [[ -n "$new_host_name" && -n "$new_ip_address" ]]; then
+if [ -n "$new_host_name" -a -n "$new_ip_address" ]; then
   # Delete entries with new_host_name or new_ip_address in /etc/hosts.
   sed -i '/Added by Google/d' /etc/hosts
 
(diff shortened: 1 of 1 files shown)


malcolmr:
I've tested this on Debian Wheezy and Jessie by applying the patch directly to `/usr/share/google/set-hostname`.  I haven't tried anything else.

malcolmr:
Incidentally, to verify this in Jessie: the dhclient output (including the error messages) is visible in the systemd journal; `sudo journalctl` will show that.

google-startup-scripts/usr/share/google/set-hostname
```diff
@@ -15,7 +14,7 @@
 
 # Deal with a new hostname assignment.
 
-if [[ -n "$new_host_name" && -n "$new_ip_address" ]]; then
+if [ -n "$new_host_name" -a -n "$new_ip_address" ]; then
```
jkaplowitz: If we're trying for maximum portability, could you do ```[ -n "$new_host_name" ] && [ -n "$new_ip_address" ]``` instead of -a (and test again)? The -a flag is an XSI extension according to SUSV3: http://pubs.opengroup.org/onlinepubs/009695399/utilities/test.html :)

malcolmr: Done; see af52001.

jkaplowitz:
Looks good, thanks! Merging.

PR #152, opened 2015-03-07
Check for a connection to the metadata server before starting Google startup scripts for Linux .

zmarano:
If a VM starts without an active network or DHCP takes too long to return an address, Google boot scripts would fail to complete.

Add a check at the very beginning of the onboot script to check for a connection to the metadata server indefinitely until it is reachable and print messages to the console when a connection cannot be obtained.

Change retry logic in get_metadata_values in cases where there are connection failures or host lookup failures and default to a 30 second retry in both cases (if for instance the metadata server is not reachable for a short period of time).

Change start on stanza's in upstart scripts to wait for started networking which should circumvent such problems on upstart based distros. The systemd scripts already do this.

### DIFF
--- a/google-startup-scripts/usr/share/google/first-boot
+++ b/google-startup-scripts/usr/share/google/first-boot
@@ -29,8 +29,7 @@ function log() {
 }
 
 function get_instance_id() {
-  # Do not retry to get the metadata value on failures.
-  MDS_TRIES=1 ${PREFIX}/usr/share/google/get_metadata_value id 2>/dev/null
+  ${PREFIX}/usr/share/google/get_metadata_value id 2>/dev/null
 }
 
 # Output the instance id.
--- a/google-startup-scripts/usr/share/google/get_metadata_value
+++ b/google-startup-scripts/usr/share/google/get_metadata_value
@@ -70,6 +70,8 @@ function get_metadata_value_with_retries() {
     case $return_code in
       # No error.  We're done.
       0) exit ${return_code};;
+      # Failed to resolve host, wait for network interfaces.  Retry.
+      6) sleep 0.3; continue;;
       # Failed to connect to host.  Retry.
       7) sleep 0.1; continue;;
       # A genuine error.  Exit.
--- a/google-startup-scripts/usr/share/google/get_metadata_value
+++ b/google-startup-scripts/usr/share/google/get_metadata_value
@@ -18,17 +18,13 @@ declare -r VARNAME=$1
 declare -r MDS_PREFIX=http://metadata.google.internal/computeMetadata/v1
 declare -r MDS_TRIES=${MDS_TRIES:-100}
 
-function metadata_get_url_code() {
-  return $(curl "${1}" -H "Metadata-Flavor: Google" -s \
-      -w "%{http_code}" -o /dev/null)
-}
-
 function print_metadata_value() {
   local readonly tmpfile=$(mktemp)
-  curl -f "${1}" -H "Metadata-Flavor: Google" > ${tmpfile} 2>/dev/null
+  http_code=$(curl -f "${1}" -H "Metadata-Flavor: Google" -w "%{http_code}" \
+    -s -o ${tmpfile} 2>/dev/null)
   local readonly return_code=$?
   # If the command completed successfully, print the metadata value to stdout.
-  if [[ ${return_code} == 0 ]]; then
+  if [[ ${return_code} == 0 && ${http_code} == 200 ]]; then
     cat ${tmpfile}
   fi
   rm -f ${tmpfile}
@@ -38,13 +34,8 @@ function print_metadata_value() {
 function print_metadata_value_if_exists() {
   local return_code=1
   local readonly url=$1
-  metadata_get_url_code ${url}
-  http_code=$?
-  # Test the instance metadata value.
-  if test ${http_code} -eq 200; then
-    print_metadata_value ${url}
-    return_code=$?
-  fi
+  print_metadata_value ${url}
+  return_code=$?
   return ${return_code}
 }
 
--- a/google-startup-scripts/usr/share/google/get_metadata_value
+++ b/google-startup-scripts/usr/share/google/get_metadata_value
@@ -14,10 +14,21 @@
 # limitations under the License.
 
 # Get a metadata value from the metadata server.
+declare -r LOGFILE=/var/log/google.log
+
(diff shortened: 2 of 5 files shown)


google-startup-scripts/usr/share/google/get_metadata_value
```diff
@@ -14,9 +14,8 @@
 # limitations under the License.
 
 # Get a metadata value from the metadata server.
-
 declare -r VARNAME=$1
-declare -r MDS_PREFIX=http://169.254.169.254/computeMetadata/v1
+declare -r MDS_PREFIX=http://metadata/computeMetadata/v1
```
jkaplowitz: We should not use the short form ```metadata```. If you want to switch away from the IP address, use ```metadata.google.internal```.

google-startup-scripts/usr/share/google/onboot
```diff
@@ -112,6 +112,41 @@ function print_ssh_key_fingerprints() {
...
+      # Failed to resolve host or connect to host.  Retry indefinitely.
+      6|7) sleep 1.0
+        log "Waiting for metadata server, attempt ${count}"
+        # After 7 minutes, add a console message denoting a probable network
+        # issue. On systems using dhclient there is an attempt to obtain an IP
+        # for 60 seconds followed by a 5 minute wait period. After 7 minutes,
+        # this cycle will have run through twice. After this period of time, it
+        # is not known when a DHCP lease might be obtained and the network
+        # interface fully operational.
+        if ((count >= 421)); then
```
jkaplowitz: I see where 421 came from, but can you make it obvious in the code? E.g. 7*60+1 with the comment making it clear this is in seconds.

google-startup-scripts/usr/share/google/onboot
```diff
@@ -112,6 +112,41 @@ function print_ssh_key_fingerprints() {
   return 0
 }
 
+function check_for_connection() {
+  local count=0
+  local return_code=0
+  local url="http://metadata/computeMetadata/v1/instance/?recursive=True"
```
jkaplowitz: Same comment about choice of hostname as on get_metadata_value line 18 - either use ```169.264.169.254``` or ```metadata.google.internal``` but not ```metadata```.

Also, it would be great not to have to repeat this URL in multiple places. Do we have a reason to reimplement get_metadata_value in onboot? If get_metadata_value is not already general enough to avoid this, can we easily make it so?

google-startup-scripts/usr/share/google/onboot
```diff
@@ -112,6 +112,41 @@ function print_ssh_key_fingerprints() {
...
+        continue;;
+      # A genuine error but a connection exists.
+      *) log "Extraneous error getting metadata ${return_code}"; break;;
+    esac
+  done
+  # Return the last return code we got.
+  return ${return_code}
+}
+
+check_for_connection
```
jkaplowitz: This should probably happen after set_interrupts and virtionet_irq_affinity, which are performance tweaks for disk and network and which don't hit the network at all (only in-guest sources like /sys and /proc). I agree this should precede first_boot since that hits the metadata server.

zmarano:
I addressed all your comments. There was actually a problem with the way get_metadata_value was working. It was doing two curl requests and the first one was doing unchecked and would fail if a connection did not exist. I merged their functionality into one call instead. PTAL

jkaplowitz:
LGTM if this works on our supported OSes.

google-startup-scripts/usr/share/google/first-boot
```diff
@@ -29,8 +29,7 @@ function log() {
 }
 
 function get_instance_id() {
-  # Do not retry to get the metadata value on failures.
-  MDS_TRIES=1 ${PREFIX}/usr/share/google/get_metadata_value id 2>/dev/null
+  ${PREFIX}/usr/share/google/get_metadata_value id 2>/dev/null
```
illfelder: Please comment on what is happening here.

google-startup-scripts/usr/share/google/get_metadata_value
```diff
@@ -14,22 +14,17 @@
 # limitations under the License.
 
 # Get a metadata value from the metadata server.
-
 declare -r VARNAME=$1
-declare -r MDS_PREFIX=http://169.254.169.254/computeMetadata/v1
+declare -r MDS_PREFIX=http://metadata.google.internal/computeMetadata/v1
```
illfelder: This should probably be https (and change everywhere else that is http).

google-startup-scripts/usr/share/google/get_metadata_value
```diff
@@ -70,8 +60,8 @@ function get_metadata_value_with_retries() {
     case $return_code in
       # No error.  We're done.
       0) exit ${return_code};;
-      # Failed to connect to host.  Retry.
-      7) sleep 0.1; continue;;
+      # Failed to resolve host or connect to host.  Retry.
+      6|7) sleep 0.3; continue;;
```
illfelder: For sanity, please document below the comment above what these code mean.

google-startup-scripts/usr/share/google/onboot
```diff
@@ -112,8 +112,42 @@ function print_ssh_key_fingerprints() {
...
+        # for 60 seconds followed by a 5 minute wait period. After 7 minutes,
+        # this cycle will have run through twice. After this period of time, it
+        # is not known when a DHCP lease might be obtained and the network
+        # interface fully operational.
+        if ((count >= 7*60+1)); then
+          log "There is likely a problem with the network."
+        fi
+        continue;;
+      # A genuine error but a connection exists.
+      *) log "Extraneous error getting metadata ${return_code}"; break;;
```
illfelder: log "Check for connection error getting metadata ${return_code}"; break;;

illfelder:
Pending a couple comment/logging changes, LGTM with the same caveat as jkaplowitz.

google-startup-scripts/usr/share/google/first-boot
zmarano: In cases where the metadata server may not be reachable, the now default behavior is to retry for 30 seconds. Setting MDS_TRIES=1 when calling get_metadata_value would override this retry and only try once. It doesn't make sense to not retry this metadata check on first-boot anymore than it would make sense in any of the other boot scripts for normal use cases. The great majority of the time, especially with the new check for a connection on boot, a retry would almost never occur. The default retry is therefore to catch any transient failures in getting metadata.

PR #156, opened 2015-04-02
Google clock sync manager

zmarano:
Google clock sync manager to sync the system clock when a migration occurs. Fix white space in existing daemons to be consistent.


jkaplowitz:
LGTM

PR #159, opened 2015-04-13
Add install from source repository instructions

davidn:
Occasionally it is necessary to install from master, for example if somebody needs the as yet unreleased clock-sync manager. This commit adds instructions on how to do this.


zmarano:
Hi David, as per Google's open source policy we need you to sign an individual CLA before accepting pull requests on GitHub. Instructions for contributing are here: https://github.com/GoogleCloudPlatform/compute-image-packages/blob/master/CONTRIB.md

PR #160, opened 2015-04-14
Use /bin/ip instead of /sbin/ip

adityakali:
Most distros have moved the 'ip' binary under /bin/ and /sbin/ip simply symlinks to it. In some cases (coreos) /sbin/ip doesn't even exist. So make address_manager use /bin/ip.


adityakali:
I have updated the commit to switch to using /bin/ip only if /sbin/ip doesn't exist.
PTAL.

zmarano:
LGTM

rlarocque2:
Why specify the path at all?  Is there any drawback to relying on $PATH?

PR #161, opened 2015-04-19
Update safe_format_and_mount

vanessabodard:
Script deployed on multiple OS, so not specific to redhat.
Before this commit, two tests are made on a non-existent file ; resulting is an error displayed in stdout.


illfelder:
Looks good to me. I'll merge it in.

Thanks!

PR #163, opened 2015-05-08
Add generate package scripts.

liubonan:
Use fpm to generate rpm or deb packages for compute image packages.


liubonan:
The current scripts in the pull request is not a right direction to generate packages. I will have another pull request later.

PR #164, opened 2015-05-21
Remove temp files when move operation fails

aulanov:
To write authorized_keys files into users' home directories
account manager first creates a temp file in /tmp and then moves
that file. Move operation can fail for various reasons (e.g.
read-only filesystem, out of space, etc). When this happens account
manager will leave stale temp files.
This change fixes this by unconditionally removing the temp file
after the move attempt.


wrigri:
This seems like an OK change.  We need a CLA in order to merge in any pull requests:
https://github.com/GoogleCloudPlatform/compute-image-packages/blob/master/CONTRIB.md

wrigri:
(Nevermind the CLA thing, now that we've determined who you are...)

PR #182, opened 2015-09-14
Adding /usr/local/bin to path before gsutil call for sudo-ing Redhat users

null0000:
User on redhat had an issue where "gcimagebundle -b <bucket>" would fail with a misleading error when run from the prompt with "sudo bash" on Redhat (works fine on Debian, didn't try with other OS's).

This adds gsutil's default location to PATH to avoid future confusion.


null0000:
Indentation should be good now

illfelder:
LGTM.

PR #183, opened 2015-09-15
undoing bad indentation change

null0000:
Indentation change caused error on undefined variable output_bucket.


null0000:
Removed version bump from PR

PR #189, opened 2015-09-22
Fixed rerun logic and exit handling in first-boot.

illfelder:


google-startup-scripts/usr/share/google/first-boot
```diff
@@ -32,41 +31,28 @@ function get_instance_id() {
...
-  return $?
-}
+declare -r INSTANCE_ID=$(get_instance_id)
 
-sync_new_instance_check
-if [[ $? != 0 ]]; then
-  # The instance hasn't changed.  Just exit.
-  exit 0
-fi
+# Checks whether the instance has changed.
```
null0000: Might help to add a note here that the function exits if the instance has already been booted.

null0000:
LGTM

google-startup-scripts/usr/share/google/first-boot
illfelder: Done.

zmarano:
LGTM

PR #191, opened 2015-11-14
Overhaul of the desired accounts script.

illfelder:
The change should improve performance by reducing the number of hanging
requests for metadata.

### DIFF
--- a/google-daemon/usr/share/google/google_daemon/accounts_manager.py
+++ b/google-daemon/usr/share/google/google_daemon/accounts_manager.py
@@ -62,10 +62,9 @@ class AccountsManager(object):
         if json_tags:
           etags = json.loads(json_tags)
           if etags:
-            self.desired_accounts.attributes_etag = etags[0]
-            self.desired_accounts.instance_sshkeys_etag = etags[1]
+            self.desired_accounts.etag = etags[0]
         reader.close()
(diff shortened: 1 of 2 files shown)


google-daemon/usr/share/google/google_daemon/accounts_manager.py
```diff
@@ -80,9 +79,7 @@ def Main(self):
           time.sleep(5)
 
         # Write the etag to pass to parent
-        json_tags = json.dumps(
-            [self.desired_accounts.attributes_etag,
-             self.desired_accounts.instance_sshkeys_etag])
+        json_tags = json.dumps([self.desired_accounts.etag])
```
EricSL: Any reason to still send this as a list?

illfelder: Possibly - if a VM upgrades to this change and that file already exists on the machine, I don't want the script to break.

EricSL: It doesn't look to me like it is ever saved to a file -- it is sent over a pipe from the forked child to the parent, so the two ends of the pipe should always be running the same binary.

I guess it's fine either way. LGTM.

illfelder: Mind taking a look at the new version of this file?

zmarano:
LGTM

PR #192, opened 2015-11-18
Add 120s timeout so Ubuntu 15+ doesn't timeout on shutdown scripts ea…

null0000:
…rly after 90s

### DIFF
--- a/google-startup-scripts/usr/lib/systemd/system/google-startup-scripts.service
+++ b/google-startup-scripts/usr/lib/systemd/system/google-startup-scripts.service
@@ -9,6 +9,7 @@ ExecStart=/usr/share/google/run-startup-scripts
 ExecStop=/usr/share/google/run-shutdown-scripts
 Type=oneshot
 RemainAfterExit=true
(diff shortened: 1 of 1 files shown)


illfelder:
This is not a change we want to make. We should not limit the time a shutdown script has to run.

null0000:
We need to specify a timeout to avoid having shutdown problems on Ubuntu.
As is, it shuts down at 90 seconds because of systemd defaults. Two minutes
is the duration vms stay up after being told to shut down, so having
timeout=120 makes sense

I can try timeout=0, but it's effectively the same thing from the users
perspective, unless we have vms staying up past two minutes once the vm is
shut down.

illfelder:
A VM has no shutdown time limit when shutdown is triggered from the guest (or any other guest operations). This is also the case for reboot.

null0000:
Fair enough. I'll change it

null0000:
Fixed. Tests passing on Debian and Ubuntu

illfelder:
LGTM.

PR #200, opened 2016-01-28
New service for shutdown scripts on systemd.

illfelder:
A running startup script should not be able to prevent a shutdown script
from executing.

### DIFF
--- a/google-startup-scripts/usr/lib/systemd/system/google-shutdown-scripts.service
+++ b/google-startup-scripts/usr/lib/systemd/system/google-shutdown-scripts.service
@@ -1,15 +1,12 @@
 [Unit]
-Description=Google Compute Engine user shutdown scripts.
-After=local-fs.target network-online.target network.target
-After=google.service rsyslog.service
+Description=Google Compute Engine user shutdown scripts
+Before=shutdown.target
+DefaultDependencies=true
 Wants=local-fs.target network-online.target network.target
 
 [Service]
-ExecStart=/bin/true
-ExecStop=/usr/share/google/run-shutdown-scripts
+ExecStart=/usr/share/google/run-shutdown-scripts
 Type=oneshot
-RemainAfterExit=true
-TimeoutStopSec=0
 
 [Install]
-WantedBy=multi-user.target
+WantedBy=shutdown.target
--- a/google-startup-scripts/usr/lib/systemd/system/google-shutdown-scripts.service
+++ b/google-startup-scripts/usr/lib/systemd/system/google-shutdown-scripts.service
@@ -1,13 +1,15 @@
 [Unit]
-Description=Google Compute Engine user startup scripts.
+Description=Google Compute Engine user shutdown scripts.
 After=local-fs.target network-online.target network.target
 After=google.service rsyslog.service
 Wants=local-fs.target network-online.target network.target
 
 [Service]
-ExecStart=/usr/share/google/run-startup-scripts
+ExecStart=/bin/true
+ExecStop=/usr/share/google/run-shutdown-scripts
 Type=oneshot
 RemainAfterExit=true
+TimeoutStopSec=0
 
 [Install]
 WantedBy=multi-user.target
--- a/google-startup-scripts/usr/lib/systemd/system/google-shutdown-scripts.service
+++ b/google-startup-scripts/usr/lib/systemd/system/google-shutdown-scripts.service
@@ -1,6 +1,6 @@
 [Unit]
 Description=Google Compute Engine user shutdown scripts
-Before=shutdown.target
+Before=shutdown.target reboot.target halt.target
 DefaultDependencies=true
 Wants=local-fs.target network-online.target network.target
 
@@ -9,4 +9,4 @@ ExecStart=/usr/share/google/run-shutdown-scripts
 Type=oneshot
 
 [Install]
-WantedBy=shutdown.target
+WantedBy=shutdown.target reboot.target halt.target
--- a/google-startup-scripts/usr/lib/systemd/system/google-shutdown-scripts.service
+++ b/google-startup-scripts/usr/lib/systemd/system/google-shutdown-scripts.service
@@ -1,7 +1,7 @@
 [Unit]
 Description=Google Compute Engine user shutdown scripts
-Before=shutdown.target reboot.target halt.target
-DefaultDependencies=true
+Before=shutdown.target
+DefaultDependencies=false
 Wants=local-fs.target network-online.target network.target
 
 [Service]
@@ -9,4 +9,4 @@ ExecStart=/usr/share/google/run-shutdown-scripts
 Type=oneshot
 
 [Install]
-WantedBy=shutdown.target reboot.target halt.target
+WantedBy=shutdown.target
--- a/google-startup-scripts/usr/lib/systemd/system/google-startup-scripts.service
+++ b/google-startup-scripts/usr/lib/systemd/system/google-startup-scripts.service
@@ -1,5 +1,5 @@
 [Unit]
-Description=Google Compute Engine user startup scripts.
+Description=Google Compute Engine user startup scripts
 After=local-fs.target network-online.target network.target
 After=google.service rsyslog.service
 Wants=local-fs.target network-online.target network.target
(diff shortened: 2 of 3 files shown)


google-startup-scripts/usr/lib/systemd/system/google-shutdown-scripts.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=Google Compute Engine user shutdown scripts.
+After=local-fs.target network-online.target network.target
+After=google.service rsyslog.service
+Wants=local-fs.target network-online.target network.target
+
+[Service]
+ExecStart=/bin/true
```
zmarano: This should be:
ExecStart=/usr/share/google/run-shutdown-scripts

And you don't need ExecStop at all. This job is to run the script- it should run when the job starts not when it stops.

google-startup-scripts/usr/lib/systemd/system/google-shutdown-scripts.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=Google Compute Engine user shutdown scripts.
+After=local-fs.target network-online.target network.target
+After=google.service rsyslog.service
```
zmarano: I don't think you want google.service here as this is for shutdown and google.service will have already run long ago.

zmarano: Also you need:
Before=shutdown.target reboot.target halt.targe
To make this run on shutdown/reboot events.

google-startup-scripts/usr/lib/systemd/system/google-shutdown-scripts.service
```diff
@@ -0,0 +1,15 @@
...
+Wants=local-fs.target network-online.target network.target
+
```
zmarano: And you probably need:
DefaultDependencies=no

google-startup-scripts/usr/lib/systemd/system/google-shutdown-scripts.service
```diff
@@ -0,0 +1,15 @@
...
+[Service]
+ExecStart=/bin/true
+ExecStop=/usr/share/google/run-shutdown-scripts
+Type=oneshot
+RemainAfterExit=true
+TimeoutStopSec=0
+
+[Install]
+WantedBy=multi-user.target
```
aulanov: Should be: WantedBy=shutdown.target

google-startup-scripts/usr/lib/systemd/system/google-shutdown-scripts.service
aulanov: In fact you only need "Before=shutdown.target" (shutdown.target is always reached before reboot.target or halt.target, so no need to specify the latter two).

google-startup-scripts/usr/lib/systemd/system/google-shutdown-scripts.service
```diff
@@ -0,0 +1,15 @@
...
+After=local-fs.target network-online.target network.target
+After=google.service rsyslog.service
+Wants=local-fs.target network-online.target network.target
+
+[Service]
+ExecStart=/bin/true
+ExecStop=/usr/share/google/run-shutdown-scripts
+Type=oneshot
+RemainAfterExit=true
+TimeoutStopSec=0
```
aulanov: These two lines are not necessary.

google-startup-scripts/usr/lib/systemd/system/google-shutdown-scripts.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=Google Compute Engine user shutdown scripts.
```
aulanov: Please remove period at the end.

google-startup-scripts/usr/lib/systemd/system/google-startup-scripts.service
```diff
@@ -1,15 +1,13 @@
 [Unit]
-Description=Google Compute Engine user scripts
+Description=Google Compute Engine user startup scripts.
```
aulanov: No period at the end necessary.

illfelder: Done.

google-startup-scripts/usr/lib/systemd/system/google-shutdown-scripts.service
illfelder: Done.

google-startup-scripts/usr/lib/systemd/system/google-shutdown-scripts.service
illfelder: Replaced the After lines with Before=shutdown.target. Thanks for the suggestions!

google-startup-scripts/usr/lib/systemd/system/google-shutdown-scripts.service
illfelder: Added.

google-startup-scripts/usr/lib/systemd/system/google-shutdown-scripts.service
illfelder: This didn't work when I tried it yesterday. It might work with a change to WantedBy like Andrey mentioned.

google-startup-scripts/usr/lib/systemd/system/google-shutdown-scripts.service
illfelder: Removed.

google-startup-scripts/usr/lib/systemd/system/google-shutdown-scripts.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Google Compute Engine user shutdown scripts
+Before=shutdown.target
+DefaultDependencies=true
+Wants=local-fs.target network-online.target network.target
```
zmarano: I don't actually know if you need this Wants line. Worth testing with and without.

google-startup-scripts/usr/lib/systemd/system/google-shutdown-scripts.service
```diff
@@ -0,0 +1,12 @@
...
+Before=shutdown.target reboot.target halt.target
+DefaultDependencies=true
+Wants=local-fs.target network-online.target network.target
+
+[Service]
+ExecStart=/usr/share/google/run-shutdown-scripts
+Type=oneshot
+
+[Install]
+WantedBy=shutdown.target reboot.target halt.target
```
aulanov: As I said you don't need to mention reboot.target and halt.target here. shutdown.target precedes them both.

google-startup-scripts/usr/lib/systemd/system/google-shutdown-scripts.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Google Compute Engine user shutdown scripts
+Before=shutdown.target reboot.target halt.target
+DefaultDependencies=true
+Wants=local-fs.target network-online.target network.target
```
aulanov: "DefaultDependencies=" should be No
"Wants=" can be removed.

See iptables-store.service for a good example of a service that runs on systemd shutdown.

google-startup-scripts/usr/lib/systemd/system/google-shutdown-scripts.service
illfelder: This is a second iteration. The previous commit did not have reboot and halt and shutdown scripts were not triggered by calling stop on the instance, or by running sudo shutdown -h now in the VM.

aulanov: > were not triggered by calling stop on the instance, or by running sudo shutdown -h now in the VM

That might be because "DefaultDependencies" was set. Also was google-shutdown-scripts.service enabled at that time?

google-startup-scripts/usr/lib/systemd/system/google-shutdown-scripts.service
illfelder: Fixed the default dependencies. Wants is needed because shutdown scripts using network shouldn't have network cut out early.

google-startup-scripts/usr/lib/systemd/system/google-shutdown-scripts.service
illfelder: I've fixed default dependencies and will try with reboot and halt removed.

zmarano:
LGTM

illfelder:
Thanks for all the pointers and suggestions Zach and Andrey!

aulanov:
LGTM

PR #201, opened 2016-02-10
Update accounts daemon with revised procedure.

illfelder:
ssh-keys in instance metadata will store all SSH keys.
block-project-ssh-keys will prevent use of SSH keys in project metadata.


zmarano:
LGTM

PR #203, opened 2016-02-10
Add experimental gce-disk-expand package for CentOS 6 and RHEL 6

zmarano:


wrigri:
LGTM (with that one caveat we discussed)

wrigri:
LGTM

PR #205, opened 2016-02-12
Trigger the shutdown script via ExecStop.

illfelder:
Executing a script as ExecStart doesn't give the same ordering
guarantees. This means that syslog can stop before shutdown scripts run.


zmarano:
lgtm

PR #206, opened 2016-02-17
- Improve security for temporary key location

rjschwei:
+ For host key regeneration the temporary location is not sufficiently
    random as it only uses the process ID as unique identifier. This is to
    a certain extend predictable. Therefore a symlink can be used to over
    write the host key files even if that may not be the intention. The
    change places the temporary key files into a directory with a random
    name, thus eliminating the risk.


zmarano:
LGTM

PR #208, opened 2016-02-17
fix url for fetching instance id

adityakali:
The instance-id is accessible from URL "http://metadata.google.internal/computeMetadata/v1/instance/id" and not "http://metadata.google.internal/computeMetadata/v1/instance/instance-id".


zmarano:
LGTM

PR #213, opened 2016-03-23
Add python3 compability to boto compute_auth.py

bivald:
Default boto configuration on Google Compute Engine is broken when using Python3 and service accounts. Boto automatically includes /etc/boto.cfg which includes compute_auth.py (which is Python2 only). This pull request makes compute_auth.py Python 2+3 safe. 

This relates to issues found on the web, such as https://github.com/travis-ci/travis-ci/issues/5246 and 
https://github.com/boto/boto/issues/3263

This is an extension of commit https://github.com/GoogleCloudPlatform/compute-image-packages/commit/d322487a971a9aa84c125bb7617a66c2bc6307d4

### DIFF
--- a/google-startup-scripts/usr/share/google/boto/boto_plugins/compute_auth.py
+++ b/google-startup-scripts/usr/share/google/boto/boto_plugins/compute_auth.py
@@ -15,7 +15,12 @@
 """Authentication module for using Google Compute service accounts."""
 
 import json
-import urllib2
+
+try:
+  from urllib2 import Request, urlopen, URLError, HTTPError
+except ImportError:
+  from urllib.request import Request, urlopen
+  from urllib.error import URLError, HTTPError
 
 from boto.auth_handler import AuthHandler
 from boto.auth_handler import NotReadyToAuthenticate
@@ -57,11 +62,11 @@ class ComputeAuth(AuthHandler):
 
   def __GetJSONMetadataValue(self, url):
     try:
-      request = urllib2.Request(url)
+      request = Request(url)
       request.add_unredirected_header('Metadata-Flavor', 'Google')
-      data = urllib2.urlopen(request).read()
+      data = urlopen(request).read()
       return json.loads(data)
-    except (urllib2.URLError, urllib2.HTTPError, IOError):
+    except (URLError, HTTPError, IOError):
       return None
 
   def __GetGSScopes(self):


zmarano:
Hi, Thanks for the PR!
Can you sign a CLA so we can accept your patch? Thanks! https://github.com/GoogleCloudPlatform/compute-image-packages/blob/master/CONTRIB.md

bivald:
Done 👍🏼

illfelder:
Hi, I am in the process of updating the entire directory to be Python 2/3 compatible along with making additional improvements. Do you need your change checked in immediately for your own development, or would you mind if we defer these changes until the larger update is made? (I want to avoid the potential for breakages in the short term if possible).

bivald:
@illfelder I would say it's better to wait for your larger updates (as you say, minimise risk for breakage). I'm already supplementing my compute_auth.py in my debian image on gce, so I'm set either way :)

illfelder:
Awesome! Thanks for the understanding! I'll close this off then.

tomsing1:
Any news on this update? I ran into it today as well when I was running VMs in GCE. Would be nice to see the provided patch applied.

illfelder:
We are in the process of overhauling the final daemon so the repository will be Python 2/3 compatible. We want to fully test the changes in new images over the next few weeks prior to publishing the changes on GitHub.

listingmirror:
@illfelder  Any update on this? This is causing a huge problem as this file gets auto populated on fresh GCE instances (no matter how many times I delete it), which crashes anything running python3. I have to do sed hackery in startup scripts to hack boto.cfg plugins directory, and then restart all the services that failed to start.

illfelder:
The new Linux guest environment is entirely Python 3 compatible and is provided by default on new Debian, CentOS, and RHEL Google Compute Engine images. Ubuntu is still updating their images to include the updated guest code.

Please let me know if you run into any issues.

listingmirror:
@illfelder Any issue to track for Python3 support in ubuntu 16?  We are an Ubuntu 16 shop, so they seem to be the one holdout from your list...

illfelder:
@OddBloke is validating the new guest environment on Ubuntu images, and would likely have a more accurate ETA.

As a workaround if you would like to use the guest environment on Ubuntu before it's officially supported/provided, the google-compute-engine-jessie package should work on Ubuntu images. Any conflicts with cloud init can be disabled using the instance defaults template file.

listingmirror:
For a current workaround I am just deleting lines out of boto.cfg in a startup script, but that seems less than ideal long term.

PR #216, opened 2016-04-20
Default users on ubuntu should be in the lxd group

OddBloke:
lxd is a core part of the Ubuntu cloud experience, so users should be in the lxd group.


OddBloke:
It sounds like this code is going to get a lot of rework soon, so we'll carry this as a distro patch for now.

Issue #218, opened 2016-04-26
IsUserSudoerEntry does not escape regex


mattjo:
The [IsUserSudoerEntry](https://github.com/GoogleCloudPlatform/compute-image-packages/blob/master/google-daemon/usr/share/google/google_daemon/accounts.py#L47) function does not escape regex characters in the user name, so if there are two users on the system, test.user and test_user, the test.user will never be added since test_user matches the regex.

illfelder:
We are rewriting the Linux accounts daemons; as part of the change, we will be adding users to a group, so weird regex parsing of users won't be necessary.

illfelder:
The rewrite is now compete and can be reviewed in the development branch of compute-image-packages.

illfelder:
The latest GitHub release contains the new Linux guest environment, and fixes this issue.

PR #219, opened 2016-04-26
Escape user in IsUserSudoerEntry re

mattjo:
The re for the filter does not escape the regular expression, so if a user name has something like a dot, this re will have an unintended match causing #218


mattjo:
I signed it!

illfelder:
We have completed a rewrite of the Linux guest environment that can be viewed in the development branch. Closing this off since it is now obsolete.

Issue #217, opened 2016-04-24
google-adress-manager is broken on Ubuntu Xenial 16.04


MatthiasWinkelmann:
address-manager fails on the current image of 16.04 server because the network interfaces have been renamed: 

```
 google-address-manager[7151]: WARNING Non-zero exit status from: "/sbin/ip route ls table local type local dev eth0 scope host proto 66"
                    STDOUT:

             STDERR:
          Cannot find device "eth0"
```

I am getting such an error every 5 seconds. See also https://bugs.launchpad.net/juju-vagrant-images/+bug/1547110 or https://github.com/geerlingguy/packer-ubuntu-1604/issues/1 for (some) of the other manifestations of this change.

israelfaria:
I did hit this too. The solution is quite simple:
Change line 110 on address_manager.py to
`cmd = ('{0} route ls table local type local dev **ens4** scope host ' +`
For some reason naming scheme of network interfaces has been changing in ubuntu. It seems to be related to systemd. Anything assuming network interface is eth0 should be revised.

MatthiasWinkelmann:
I got the impression that the network interface IDs are now subject to change without notice, which would make this a temporary solution at best... But we'll see.

zmarano:
Thanks for reporting this! We are aware of this issue and working with Canonical on a fix for Ubuntu 16.04. We also have a fix in the works to make this user or provider configurable. Stay tuned.

CrackerJackMack:
@zmarano you have a link to the launchpad bug so we can follow along?

zmarano:
Canonical has released the patch into Xenial. You should get the fixed version if you update your instance (apt-get update; apt-get dist-upgrade). Also new instances based on ubuntu-1604-xenial-v20160429 will have this fix.

sbadia:
For the record…
```
# zcat /usr/share/doc/gce-daemon/changelog.Debian.gz|head -20
gce-utils (1.3.3-0ubuntu4~16.04.0) xenial; urgency=medium

  * Backport to 16.04

 -- Daniel Watkins <daniel.watkins@canonical.com>  Tue, 26 Apr 2016 11:01:13 +0100

gce-utils (1.3.3-0ubuntu4) yakkety; urgency=medium

  * Don't assume (in address manager) that the network interface will be eth0.

 -- Daniel Watkins <daniel.watkins@canonical.com>  Mon, 25 Apr 2016 17:54:22 +0100
```

[...]

Issue #110, opened 2014-08-25
address_manager.py support for systems without iproute2


ok-devalias:
Currently, these scripts only work for Linux systems.  In particular, porting address_manager.py would allow for FreeBSD and potentially other UNIX systems to use the load balancing side of GCE.  I have been attempting to translate the "ip route add to local $IP/32 dev eth0 proto 66" command, but so far I've had no luck.  Looks like "-proto2" can replace "proto 66."

I've tried:
- route add -host $LB_IP -iface vtnet0 -nostatic -proto2
- route add -host $LB_IP $HOST_IP -nostatic -proto2

bsiegert:
Any update on this? Looking into porting this to NetBSD.

zmarano:
address_manager has been deprecated in favor of the ip_forwarding_daemon.  At this point we only support Linux in the "Linux guest environment'. However, we do accept PR's if you have a specific change in mind that would solve this elsewhere.

PR #220, opened 2016-04-29
address_manager: Remove hard-coding of « eth0 » interface

sbadia:
Added a dynamic default interface retrival, using `/proc` in order to no
add an extra dependency (`netifaces`, `pynetinfo`).

This PR, manage both predictable network interface names (`en{s,o,p}`), and old fashion interface  (`eth0`) names.

* https://www.freedesktop.org/wiki/Software/systemd/PredictableNetworkInterfaceNames/
* https://github.com/systemd/systemd/blob/master/src/udev/udev-builtin-net_id.c

Refs: #217

### DIFF
--- a/google-daemon/usr/share/google/google_daemon/address_manager.py
+++ b/google-daemon/usr/share/google/google_daemon/address_manager.py
@@ -15,8 +15,8 @@
 """Manage extra network interface addresses on a GCE instance.
 
 Fetch a list of public endpoint IPs from the metadata server, compare it with
-what's configured on eth0, and add/remove addresses from eth0 to make them
-match.  Only remove those which match our proto code.
+what's configured on default interface, and add/remove addresses from default
+interface to make them match.  Only remove those which match our proto code.
 
 This must be run by root. If it reads any malformed data, it will take no
 action.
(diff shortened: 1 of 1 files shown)


illfelder:
Hi Sebastien,

Thanks for your contribution! We just completed an overhaul of the Linux guest environment that includes large changes to the address daemon. We are still considering ways to dynamically detect the default interface so I'll try to merge your change into the updated version.

sbadia:
Hi @illfelder ,

Thanks for the comment and informations! I don't know (about the policy for external libs usage), but maybe we can manage this by removing system calls to `ip` and use a common (linux/bsd*/win*) python interface like [netifaces](https://pypi.python.org/pypi/netifaces), in order to solve both issues (default interface #220 / #217 and #110)

wdyt?

Cheers,

CrackerJackMack:
@illfelder Any chance we can see that overhaul in a PR?

sbadia:
Don't know where the patch was commited (seems not a public repository), but it's now fixed in version `1.3.3-0ubuntu4` see https://github.com/GoogleCloudPlatform/compute-image-packages/issues/217#issuecomment-215817565

zmarano:
TL;DR:
Canonical patched their branch to fix Ubuntu 16.04 immediately because it was breaking users. We are working on releasing our rewrite today in a separate branch here (stay tuned). We'll pull in your patch to that branch as well. We won't merge that new branch with master until we are happy with testing however.

illfelder:
The code is now pushed to a development branch of compute image packages:
https://github.com/GoogleCloudPlatform/compute-image-packages/tree/development

sbadia:
Thanks!

PR #228, opened 2016-05-04
Move instance setup scripts to usr/bin.

illfelder:
usr/local/bin does not appear on the path in certain distros.


zmarano:
LGTM

PR #229, opened 2016-05-04
Separate out the testing compatibility file.

illfelder:
- Testing compatibility includes dependencies not present in base images.
- Includes a fix for usermod ordering needed by older distros.


zmarano:
LGTM

PR #230, opened 2016-05-04
Account and IP forwarding compatibility fixes.

illfelder:
- Default interface should be the value of dev.
- User should be the last parameter of user account commands.
- All files created should have restorecon run on them.


zmarano:
LGTM

PR #232, opened 2016-05-10
Generate instance defaults during instance setup.

illfelder:
The instance configuration file is based on a user created template.
Users modify defaults via the template configuration file. Instance
setup will respect user defaults and not override them. If a template
does not exist, instance setup will generate a file with Google
defaults.


zmarano:
LGTM

PR #233, opened 2016-05-11
Improved locking in the config manager library.

illfelder:
Removed a global lock when writing to different config files.


zmarano:
LGTM

PR #240, opened 2016-05-18
Style improvements to the Linux packaging.

illfelder:
- Prefix release number with "0." so distros can update and pin to a release.
- Replace underscores with hyphens in package name.
- Add packages to dist-packages instead of sites-packages in deb. This is
consistent with debs in Debian and Ubuntu. Adding to dist-packages avoids
conflicting with Python installed from source.


zmarano:
LGTM

PR #242, opened 2016-05-19
Fix packaging bugs.

zmarano:
- Fix entry points and add main functions in daemon's and instance_setup.
- Fix systemd post scripts.
- Fix fpm python data directory (which gets created in the pkg).


illfelder:
LGTM.

PR #243, opened 2016-05-20
Fix packaging on sysvinit.

illfelder:
- Main functions used as endpoints can't accept arguments.
- Add package dependencies.
- Remove version dependency in Python package.


zmarano:
LGTM

PR #244, opened 2016-05-20
Improve and fix the sysvinit guest configs.

illfelder:
- Shutdown scripts should execute on VM shutdown.
- Logging is handled by daemons instead of the sysvinit configs.
- Logger module optionally does not write to the console.
- Config file for writing logs to serial port output.


zmarano:
LGTM

PR #246, opened 2016-05-25
Remove auth SysLog from logging to the console.

illfelder:
- The new Linux guest environment does not write to the auth facility.
- Removing auth will prevent sshd spam from malicious SSH attempts.
- Improvements to the script for building the configuration packages.


zmarano:
LGTM

PR #247, opened 2016-05-25
Fix style in build_packages

zmarano:


illfelder:
LGTM. End the commit message with a period plox.

PR #249, opened 2016-05-28
Fix logging and a bug in instance config.

illfelder:
- The instance config should reuse the instance config file if it exists.
- The logger should not propagate message to the system logger.
- Remove duplicate logging to console from a running daemon.
- Add a debug mode for executing scripts and daemons.

More debug logging can be added as appropriate.


zmarano:
LGTM

PR #250, opened 2016-06-01
Change dependency to python-setuptools and distro names to jessie and wheezy for Debian.

zmarano:


illfelder:
Commit titles should be <= 50 chars.

illfelder:
LGTM!

PR #252, opened 2016-06-01
Create the base directory before copying files.

illfelder:
The shutil copy command will raise an exception when the base directory
for a file does not exist.


zmarano:
LGTM

PR #254, opened 2016-06-02
Add a default config file for partner defaults.

illfelder:
This enables a partner to provide their own default values without
overriding user settings during package upgrades.


zmarano:
LGTM

zmarano:
LGTM

PR #273, opened 2016-06-17
[build] support ubuntu 14.04

dmichel1:
Note that the google-config package conflicts with the  files gcp-startup used to provide, specially 90-google.conf and 11-gce-network-security.conf. This is assuming the gcp-startup package is being deprecated in favor of splitting out the startup sripts and configs (btw big +1).

It would be great to have these packages uploaded to the cloud package repo.

### DIFF
--- a/build_packages.sh
+++ b/build_packages.sh
@@ -55,3 +55,6 @@ build_distro 'el7' 'rpm' 'systemd' '/usr/lib/python2.7/site-packages'
 # Debian
 build_distro 'wheezy' 'deb' 'sysvinit' '/usr/lib/python2.7/dist-packages'
 build_distro 'jessie' 'deb' 'systemd' '/usr/lib/python2.7/dist-packages'
+
+# Ubuntu
+build_distro 'ubuntu-14.04' 'deb' 'upstart' '/usr/lib/python2.7/dist-packages'
--- a/google_configs/build_packages.sh
+++ b/google_configs/build_packages.sh
@@ -42,23 +42,24 @@ function build_distro() {
     --url 'https://github.com/GoogleCloudPlatform/compute-image-packages' \
     --vendor 'Google Compute Engine Team' \
     --version '2.0.0' \
+    --replaces 'gce-startup-scripts' \
     "${COMMON_FILES[@]}" \
     "${files[@]:2}"
 }
 
-# RHEL/CentOS 6
+# RHEL/CentOS
 build_distro 'el6' 'rpm' \
   'bin/set_hostname=/etc/dhcp/dhclient-exit-hooks'
-
-# RHEL/CentOS 7
 build_distro 'el7' 'rpm' \
   'bin/set_hostname=/usr/bin/set_hostname' \
   'dhcp/google_hostname.sh=/etc/dhcp/dhclient.d/google_hostname.sh'
 
-# Debian 7
+# Debian
 build_distro 'wheezy' 'deb' \
   'bin/set_hostname=/etc/dhcp/dhclient-exit-hooks.d/set_hostname'
-
-# Debian 8
 build_distro 'jessie' 'deb' \
+  'bin/set_hostname=/etc/dhcp/dhclient-exit-hooks.d/set_hostname'
+
+# Ubuntu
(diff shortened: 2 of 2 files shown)


dmichel1:
I signed it! Spotify AB is the corporate agreement.

illfelder:
Hi Drew,

Thanks for your contribution! A few things to mention:

*   We don't currently include Ubuntu as a package config option because there are setting changes specific to Ubuntu images. In particular, we will need to adjust [instance_defaults.cfg](https://github.com/GoogleCloudPlatform/compute-image-packages/blob/master/package/instance_configs.cfg) to disable parts of instance setup that should not execute. We are actively working on supporting the new guest in Ubuntu images.

*   The new guest does replace the old packages, but gce-startup-scripts is the Ubuntu specific package name (the package is google-startup-scripts on other Linux distros). I like the idea of adding a replacement to the fpm command, but I want to make sure the change will work across various distros.

*   The (non-Ubuntu) packages are hosted and available in Google's cloud package repository. Packaging instructions can be found in the [package distribution](https://github.com/GoogleCloudPlatform/compute-image-packages#package-distribution) section of our README.

*   We've changed our contribution process where all pull requests should be made against the development branch. The change is mentioned in the [contributing](https://github.com/GoogleCloudPlatform/compute-image-packages#contributing) section of our README.

dmichel1:
Hi Max,

* That is great news! Any idea on when Ubuntu support would be completed? I'm trying to weigh the options of building this internally vs waiting for the official Ubuntu packages. I was going to open an issue about set_multiqueue failing to apply (didn't have time to dig into it yet). Is that the bit that needs to be disabled/fixed in `instance_defaults`?

* I can modify the replacement bit to be specific to Ubuntu while giving the other distros the option to specify the relevant package names to replace. There's also a similar issue between gce-daemon <--> google-compute-engine where the old daemons will be left running.

* Ack, I missed that part and will re-open any PR's against development 👍

illfelder:
We are working with Canonical on making the changes, but I unfortunately don't have a firm timeline at the moment. Significant validation is likely necessary to verify the changes on Ubuntu images and to add the new guest into their images. I don't know offhand if multiqueue support is one of the problematic components, so if you have the error message, I'd be interested in taking a look. Some of the functionality we offer (e.g. setting host keys) is not needed due to image configurations or cloud init, so we expect the Canonical package will have the functionality disabled by default.

Awesome (with respect to the other bullet points)!

dmichel1:
We don't use the Ubuntu images provided by Google.

To give you some context into my environment we build GCE images off our internal image that we use for bare-metal installations. It's essentially a tar-ball containing a run of `debootstrap minimum` for Ubuntu 14.04 and a run of puppet which configures the rest of the OS. That tar-ball is then pulled into our GCE image build process where we apply Google specific configurations. 

I'm actually looking into removing cloud-init as a dependency in our images, as the new guest  specifically the `google-compute-engine` package does 90% of the job (startup/shutdown scripts, ssh keys, etc). The only requirement missing for us is disk resizing during boot which we can work around. 

I would like to have the guest packages uploaded for users of Ubuntu who build their own images and leave it up to them how they want to configure it. 

Side Note - Do you have any documentation for what changed between the old and new packages (e.g. get_metadata_value which is provided by gce-startup-scripts is missing in the new guest)?

illfelder:
Thanks for the information - it's very helpful. I'll discuss with other team members about the best path forward regarding Ubuntu packaging. One of my immediate concerns is causing confusion if Ubuntu releases their own package.

Regarding changes, the README describes the entirety of how the new Linux guest environment works. One of the largest changes is that we have created a common library for retrieving all data from the metadata server (entirely written in Python 2.6 - 3.5 version compatible). The [Metadata Watcher](https://github.com/illfelder/compute-image-packages#metadata-watcher) provides a library to retrieve the contents of the metadata server.

dmichel1:
I have some bash scripts that can't use the python library directly and that's where `get_metadata_value` was used. Any plans to add something like that back to the guest? I can work on that bit as well.

The new README is great, nice job on that. However what I'm looking for is more of a migration guide from the old packages to the new.

PR #280, opened 2016-06-21
The Python package should not install data files.

illfelder:
All init files are installed via the google_configs package.


google_configs/build_packages.sh
```diff
@@ -23,42 +24,56 @@ TIMESTAMP="$(date +%s)"
 function build_distro() {
   declare -r distro="$1"
   declare -r pkg_type="$2"
+  declare -r init_config="$3"
+  declare -r init_prefix="$4"
   declare files=("$@")
+  declare file_pattern='*[^.sh]'
+  declare init_files=(${init_config}/${file_pattern})
   declare name='google-config'
```
zmarano: declare depends='google-compute-engine'

google_configs/build_packages.sh
```diff
@@ -23,42 +24,56 @@ TIMESTAMP="$(date +%s)"
...
 
+  for ((i=0; i<${#init_files[@]}; i++)); do
+    file_name="$(basename ${init_files[$i]})"
+    file_entry="${init_config}/${file_name}=${init_prefix}/${file_name}"
+    init_files[$i]="$file_entry"
+  done
+
   if [[ "${pkg_type}" == 'deb' ]]; then
     name="${name}-${distro}"
   fi
```
zmarano: depends="google-compute-engine-${distro}"

google_configs/build_packages.sh
```diff
@@ -23,42 +24,56 @@ TIMESTAMP="$(date +%s)"
...
 
   fpm \
     -s dir \
     -t "${pkg_type}" \
-    --description 'Google Compute Engine guest configs' \
+    --after-install "${init_config}/postinst.sh" \
+    --before-remove "${init_config}/prerm.sh" \
+    --depends "google-compute-engine-${distro}" \
```
zmarano: --depends "${depends}" \

google_configs/build_packages.sh
```diff
@@ -23,42 +24,56 @@ TIMESTAMP="$(date +%s)"
...
     --iteration "0.${TIMESTAMP}" \
     --license 'Apache Software License' \
     --maintainer 'gc-team@google.com' \
     --name "${name}" \
+    --replaces 'gce-startup-scripts' \
+    --replaces 'google-startup-scripts' \
     --rpm-dist "${distro}" \
     --url 'https://github.com/GoogleCloudPlatform/compute-image-packages' \
     --vendor 'Google Compute Engine Team' \
     --version '2.0.0' \
```
zmarano: 2.0.1 (or 2.1.0) here and in setup.py

illfelder: The version bump should be a separate commit when we are ready to cut a release.

google_configs/build_packages.sh
illfelder: Done.

google_configs/build_packages.sh
illfelder: Thanks.

dmichel1:
FPM provides mechanisms to have the init scripts installed as part of the package rather than do it as part of setup.py. I don't think there should be a dependency between google-configs and google-compute-engine. google-compute-engine should be able to stand alone and has different concerns.

```
--deb-init FILEPATH           (deb only) Add FILEPATH as an init script
--rpm-init FILEPATH           (rpm only) Add FILEPATH as an init script
```

zmarano:
Normally I would agree- and at one point we tried that. But FPM didn't have mechanisms to deal with init scripts for systemd and upstart and sysvinit across the distros we support correctly. It didn't end up putting things in the right places. I agree with the sentiment however.

jonparrott:
I'm okay with the os package for the python package having the scripts, but would in general prefer them to be in a separate package. So long as the python package's setup.py doesn't have to deal with the init scripts itself.

dmichel1:
The fun of packaging across distros 🎉. The idea behind FPM is awesome but if it doesn't provide the correct functionality it might be best to start writing spec and debian files? 

I'm coming at this from the perspective that the `google-configs` are optional. I would rather not have to install them to get the daemons to function properly. That might not be the actual case and there is more coupling than what I understand.

illfelder:
I'm closing this pull request in favor of a #283 where I create a separate package for initialization files.

PR #281, opened 2016-06-21
Updating travis to use tox

jonparrott:
Fixes #275 and #276 and starts on #274.

@illfelder PTAL

PR #282, opened 2016-06-21
Updating travis to use tox

jonparrott:


jonparrott:
auto-merging based on previous review in #281

PR #283, opened 2016-06-22
Create a Google Compute Engine init package.

### LINKED ISSUE #277: Split init scripts into separate OS packages
As per discussion with @illfelder, having `setup.py` handle installing the init scripts is not cool from a packaging perspective. If possible, it'd be better to have a separate OS package for the init scripts.

illfelder:
Fixes #277 

The google-compute-engine package is identical to the Python package and
does not install distribution specific init files. The new package is
necessary to automatically run daemons and instance setup.


zmarano:
LGTM

illfelder:
@dmichel1 - please take a look. I'd like your input before I merge.

illfelder:
Merging - if there are additional comments we can resolve them in a followup pull request.

dmichel1:
LGTM +1

PR #286, opened 2016-06-23
Support distro configs for instance defaults.

illfelder:
This allows image providers and users to specify instance defaults without conflicting.


zmarano:
LGTM

PR #288, opened 2016-06-24
Update exception handling in the metadata watcher.

illfelder:
* An error message should only be printed once to prevent log spam.
* Calls to the metadata watcher will retry until metadata is retrieved.

### DIFF
--- a/google_compute_engine/metadata_watcher.py
+++ b/google_compute_engine/metadata_watcher.py
@@ -140,6 +140,30 @@ class MetadataWatcher(object):
         break
     return json.loads(response.read().decode('utf-8'))
 
+  def _HandleMetadataUpdate(self, metadata_key='', recursive=True, wait=True):
+    """Wait for a successful metadata response.
+
+    Args:
+      metadata_key: string, the metadata key to watch for changes.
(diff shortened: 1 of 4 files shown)


google_compute_engine/metadata_watcher.py
```diff
@@ -140,6 +140,30 @@ def _GetMetadataUpdate(self, metadata_key='', recursive=True, wait=True):
...
+      try:
+        return self._GetMetadataUpdate(
+            metadata_key=metadata_key, recursive=recursive, wait=wait)
+      except (httpclient.HTTPException, socket.error, urlerror.URLError) as e:
+        if isinstance(e, type(exception)) and e.args == exception.args:
+          continue
+        else:
+          exception = e
+          message = 'GET request error retrieving metadata. %s.'
+          self.logger.error(message, str(exception))
```
jonparrott: why `str(exception)`, why not just `exc_info=True`?

jonparrott: Or even better, since you're doing `.error` just do `.exception`.

illfelder: Thanks!

zmarano:
LGTM

PR #293, opened 2016-07-07
Add a null handler to GCE loggers

thobrla:
This fixes an issue where the compute auth boto plugin would
emit a warning message to the root logger that no handlers
could be found for the "compute-auth" logger when a
GCE service account was not configured.

### DIFF
--- a/google_compute_engine/compat.py
+++ b/google_compute_engine/compat.py
@@ -15,13 +15,13 @@
 
 """A module for resolving compatibility issues between Python 2 and Python 3."""
 
+import logging
 import sys
 
 if sys.version_info >= (3,):
   # Python 3 imports.
   import configparser as parser
   import http.client as httpclient
-  from logging import NullHandler
   import urllib.error as urlerror
   import urllib.parse as urlparse
   import urllib.request as urlrequest
@@ -30,22 +30,21 @@ else:
   # Python 2 imports.
   import ConfigParser as parser
   import httplib as httpclient
-
-  import logging
-  try:  # Python 2.7+
-      from logging import NullHandler
-  except ImportError:
-      class NullHandler(logging.Handler):
-          def emit(self, record):
-              pass
-
-          def handle(self, record):
-              pass
-
-          def createLock(self):
-              pass
-
   import urllib as urlparse
   import urllib as urlretrieve
   import urllib2 as urlrequest
   import urllib2 as urlerror
+
+if sys.version_info < (2,7,):
+  class NullHandler(logging.Handler):
+    def emit(self, record):
+      pass
+
+    def handle(self, record):
+      pass
+
+    def createLock(self):
+      pass
+
+  logging.NullHandler = NullHandler
+
--- a/google_compute_engine/compat.py
+++ b/google_compute_engine/compat.py
@@ -47,4 +47,3 @@ if sys.version_info < (2,7,):
       pass
 
   logging.NullHandler = NullHandler
-
--- a/google_compute_engine/compat.py
+++ b/google_compute_engine/compat.py
@@ -21,6 +21,7 @@ if sys.version_info >= (3,):
   # Python 3 imports.
   import configparser as parser
   import http.client as httpclient
+  from logging import NullHandler
   import urllib.error as urlerror
   import urllib.parse as urlparse
   import urllib.request as urlrequest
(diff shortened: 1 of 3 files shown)


google_compute_engine/tests/logger_test.py
```diff
@@ -24,27 +24,31 @@ class LoggerTest(unittest.TestCase):
...
     # Verify logger setup with a facility.
     address = '/dev/log'
     facility = 1
     named_logger = logger.Logger(name=name, debug=True, facility=facility)
     mock_syslog.assert_called_once_with(address=address, facility=facility)
     mock_syslog.setLevel.assert_called_once_with(logger.logging.INFO)
-    self.assertEqual(named_logger.handlers, [mock_stream, mock_syslog])
+    self.assertEqual(
+        named_logger.handlers,
+        [mock_null, mock_stream, mock_syslog])
```
illfelder: Please move this up to the line above.

google_compute_engine/logger.py
```diff
@@ -32,6 +32,7 @@ def Logger(name, debug=False, facility=None):
   """
   logger = logging.getLogger(name)
   logger.handlers = []
+  logger.addHandler(logging.NullHandler())
```
jonparrott: @illfelder does this library need to support 2.6? NullHandler isn't available pre-2.7.

See: https://github.com/google/google-api-python-client/blob/master/googleapiclient/__init__.py#L20

illfelder:
Looks good other than the nit style comment. I'll merge after the comment is addressed and the tests pass.

google_compute_engine/logger.py
illfelder: Thanks for catching that! The library does need to support 2.6 (I expect the 2.6 test to fail).

illfelder: You can add the null handler import here:
https://github.com/GoogleCloudPlatform/compute-image-packages/blob/master/google_compute_engine/compat.py

google_compute_engine/compat.py
```diff
@@ -21,6 +21,7 @@
   # Python 3 imports.
   import configparser as parser
   import http.client as httpclient
+  from logging import NullHandler
```
illfelder: This can just be import logging.

illfelder:
```
import logging
import sys

if sys.version_info >= (3,):
  # Python 3 imports.
  import configparser as parser
  import http.client as httpclient
  import urllib.error as urlerror
  import urllib.parse as urlparse
  import urllib.request as urlrequest
  import urllib.request as urlretrieve
else:
  # Python 2 imports.
  import ConfigParser as parser
  import httplib as httpclient
  import urllib as urlparse
  import urllib as urlretrieve
  import urllib2 as urlrequest
  import urllib2 as urlerror


class NullHandler(logging.Handler):
  def emit(self, record):
    pass

  def handle(self, record):
    pass

  def createLock(self):
    pass


if sys.version_info < (2,7,):
  logging.NullHandler = NullHandler
```

google_compute_engine/logger.py
```diff
@@ -18,7 +18,7 @@
 import logging
 import logging.handlers
```
illfelder: Logging is getting overridden - you should be able to exclusively use the logging module from compat.

google_compute_engine/compat.py
```diff
@@ -30,22 +30,21 @@
...
+      pass
+
+    def handle(self, record):
+      pass
+
+    def createLock(self):
+      pass
+
+  logging.NullHandler = NullHandler
+
```
illfelder: Please remove the trailing blank line.

google_compute_engine/compat.py
illfelder: On second thought (and looking at your later iteration) how about this:

```
import logging
import sys

if sys.version_info >= (3,):
  # Python 3 imports.
  import configparser as parser
  import http.client as httpclient
  import urllib.error as urlerror
  import urllib.parse as urlparse
  import urllib.request as urlrequest
  import urllib.request as urlretrieve
else:
  # Python 2 imports.
  import ConfigParser as parser
  import httplib as httpclient
  import urllib as urlparse
  import urllib as urlretrieve
  import urllib2 as urlrequest
  import urllib2 as urlerror


if sys.version_info >= (2,7,):
  from logging import NullHandler
else:
  class NullHandler(logging.Handler):

    def emit(self, record):
      pass

    def handle(self, record):
      pass

    def createLock(self):
      pass
```

google_compute_engine/logger.py
illfelder: I think you can remove the two lines above.

google_compute_engine/compat.py
illfelder: Remove the blank line?

illfelder:
LGTM pending the passing test suite. Thanks!

PR #297, opened 2016-07-14
Prevent compute auth from hanging.

illfelder:
When a user does not have service account credentials, the compute auth
script should give a warning instead of hanging.


thobrla:
LGTM

PR #299, opened 2016-07-18
Fix dependencies for init package. Update google-compute-engine-init to 2.0.1.

zmarano:
We need to depend on google-configs package as well to handle previous guest updates properly.


google_compute_engine_init/build_packages.sh
```diff
@@ -54,7 +57,7 @@ function build_distro() {
     --rpm-dist "${distro}" \
     --url 'https://github.com/GoogleCloudPlatform/compute-image-packages' \
     --vendor 'Google Compute Engine Team' \
-    --version '2.0.0' \
+    --version '2.0.1' \
```
illfelder: Move this into its own commit.

PR #302, opened 2016-07-19
Fix migration from existing guest environment to new guest environment.

zmarano:
Specifically, in EL7 systemd will still not be setup properly in some instances of guest env migration. Specifying a trigger to run after uninstalling the old package to re-enable the new services and run instance setup fixes this problem.

In EL6, upstart would run startup scripts again because google-instance-setup, the init job, had run. Run it manually instead of with upstart to prevent startup scripts from running.

### DIFF
--- a/google_compute_engine_init/systemd/postinst.sh
+++ b/google_compute_engine_init/systemd/postinst.sh
@@ -26,8 +26,8 @@ systemctl enable google-ip-forwarding-daemon.service
 systemctl enable google-shutdown-scripts.service
 systemctl enable google-startup-scripts.service
 
-# Run instance setup.
-systemctl start --no-block google-instance-setup
+# Run instance setup manually to prevent startup script execution.
+/usr/bin/google_instance_setup
 
 # Start daemons.
 systemctl start --no-block google-accounts-daemon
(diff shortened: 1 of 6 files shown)


google_compute_engine_init/upstart/rpm_replace
```diff
@@ -0,0 +1,10 @@
+# Replace existing guest in EL6.
+
+# Run instance setup
+/usr/bin/google_instance_setup
+
+# Manually start daemon's
+start --no-wait google-accounts-daemon
+start --no-wait google-clock-skew-daemon
+start --no-wait google-ip-forwarding-daemon
+
```
illfelder: Please remove this line.

google_compute_engine_init/systemd/postinst.sh
```diff
@@ -26,8 +26,8 @@ systemctl enable google-ip-forwarding-daemon.service
 systemctl enable google-shutdown-scripts.service
 systemctl enable google-startup-scripts.service
 
-# Run instance setup.
-systemctl start --no-block google-instance-setup
+# Run instance setup manually.
```
illfelder: Please add to your comment "to prevent startup script execution".

illfelder:
LGTM.

PR #306, opened 2016-07-21
Add user to docker group if exists.

jeffmendoza:
If the user has sudo access, they will already have access to `sudo docker`. This helps usability by removing the need to use sudo. Also, it solves problems where the `/root` directory is mounted read-only on GCI and `sudo docker` cannot save any config to `/root/.docker`

### DIFF
--- a/google_compute_engine/instance_setup/instance_config.py
+++ b/google_compute_engine/instance_setup/instance_config.py
@@ -41,7 +41,7 @@ class InstanceConfig(config_manager.ConfigManager):
   instance_config_options = {
       'Accounts': {
           'deprovision_remove': 'false',
-          'groups': 'adm,dip,lxd,plugdev,video',
+          'groups': 'adm,dip,docker,lxd,plugdev,video',
       },
       'Daemons': {
           'accounts_daemon': 'true',


google_compute_engine/accounts/accounts_utils.py
```diff
@@ -46,10 +46,11 @@ def __init__(self, logger, groups=None, remove=False):
     self.google_sudoers_file = '/etc/sudoers.d/google_sudoers'
     self.google_users_dir = '/var/lib/google'
     self.google_users_file = os.path.join(self.google_users_dir, 'google_users')
+    self.default_groups = [ self.google_sudoers_group, 'docker', ]
```
illfelder: Adding a default group belongs as part of the instance config defaults file:
https://github.com/GoogleCloudPlatform/compute-image-packages/blob/master/google_compute_engine/instance_setup/instance_config.py#L44

jeffmendoza:
Fixed, thanks!

google_compute_engine/instance_setup/instance_config.py
```diff
@@ -41,7 +41,7 @@ class InstanceConfig(config_manager.ConfigManager):
   instance_config_options = {
       'Accounts': {
           'deprovision_remove': 'false',
-          'groups': 'adm,dip,lxd,plugdev,video',
+          'groups': 'adm,dip,lxd,plugdev,video,docker',
```
illfelder: nit: mind keeping it alphabetical? Otherwise looks good.

illfelder: Also, can you make this PR against the development branch? We don't check in directly to master.

jeffmendoza:
#309

PR #309, opened 2016-07-21
Add user to docker group if exists.

jeffmendoza:
If the user has sudo access, they will already have access to sudo docker. This helps usability by removing the need to use sudo. Also, it solves problems where the /root directory is mounted read-only on GCI and sudo docker cannot save any config to /root/.docker


illfelder:
LGTM.

PR #315, opened 2016-07-29
[RFC] Add support for RHEL5

helen-fornazier:
Add support for RHEL5 in google-compute-engine, google-compute-engine-init and google-config

* Add sysvinit-rhel5 folder
* Fix pattern in build_packages script to include the startup/shutdown scripts in the build
* RHEL5 doesn't use the /etc/sudoers.d/ folder, use /etc/sudoers file if the folder doesn't exist

### DIFF
--- a/build_packages.sh
+++ b/build_packages.sh
@@ -21,6 +21,7 @@ function build_distro() {
   declare -r distro="$1"
   declare -r pkg_type="$2"
   declare -r py_path="$3"
+  declare -r py_pkt="$4"
   declare depends='google-compute-engine-init'
   declare name='google-compute-engine'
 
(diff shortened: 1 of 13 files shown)


illfelder:
Hi Helen,

There are a few reasons we don't currently want to merge this into our main GitHub repository.

- This pull request gives the impression that RHEL 5 is supported with a standard package install. The configuration needed to make this work on a vanilla RHEL 5 image is non-trivial because EL 5 uses Python 2.4 by default. There are also several features of the guest environment I do not believe would work properly using these changes.

- Changes in your pull request specific to RHEL 5 should not be used on other Linux distributions. In particular, the way the sudoers group is created is suboptimal on newer Linux distributions. Even on RHEL 5, I would suggest modifying the file differently (you can create a new file, write to it, and swap it into place).

We'll keep these changes in a separate GitHub repository so that we can continue to provide them to users who might find them helpful, but I'm closing off the pull request.

PR #322, opened 2016-08-19
Version bump the Python and init packages.

illfelder:
- Introducing a new service and common library in the Python package.
- Add init configuration to run the new service on boot.


zmarano:
LGTM

PR #335, opened 2016-09-06
Make enabling Ethernet interfaces EL compatible.

illfelder:


wrigri:
LGTM

(It may be worth some additional wording in the comments about why you don't accept a single interface as a parameter)

PR #344, opened 2016-10-03
Add a timeout option to metadata watcher calls.

illfelder:


google_compute_engine/metadata_watcher.py
```diff
@@ -127,27 +131,30 @@ def _GetMetadataUpdate(self, metadata_key='', recursive=True, wait=True):
...
-      response = self._GetMetadataRequest(metadata_url, params=params)
+      response = self._GetMetadataRequest(
+          metadata_url, params=params, timeout=timeout)
       etag_updated = self._UpdateEtag(response)
-      if wait and not etag_updated:
+      if wait and not etag_updated and not timeout:
         # Retry until the etag is updated.
         continue
       else:
         # Waiting for change is not required or the etag is updated.
```
crunk1: This comment is out of date now. It is possible for wait=True, etag_updated=False, and timeout != None. In fact, I'm not sure I follow the logic behind "if wait and not etag_updated and not timeout".

illfelder: Thanks - updated.

PR #347, opened 2016-10-14
Add setuptools as a package dependency

rjschwei:
+ The setup.py file is used to generate scripts. The generated scripts
    create a runtime dependence on the setuptools, thus these must be
    installed on the target system where the code is supposed to run.


rjschwei:
Wrong branch will sent new PR against devel

PR #348, opened 2016-10-19
Deps

rjschwei:
"binaries" are generated via setuptools this leaves a runtime dependency on setuptools, thus it needs to be a requirement.


rjschwei:
OK, that didn't work, rebase properly.

PR #349, opened 2016-10-19
No call missing exec

rjschwei:
Do not call gsutil if it is not installed, provide appropriate messaging for the user.


rjschwei:
OK, that obviously did not work, better rebase my branch properly.

PR #351, opened 2016-10-19
- Handle exception if we cannot write the sudoers file

rjschwei:


google_compute_engine/accounts/tests/accounts_utils_test.py
```diff
@@ -15,6 +15,7 @@
 
 """Unittest for accounts_utils.py module."""
 
+
```
illfelder: Please remove this newline.

google_compute_engine/accounts/tests/accounts_utils_test.py
```diff
@@ -148,6 +149,23 @@ def testCreateSudoersGroupError(
     ]
     self.assertEqual(mocks.mock_calls, expected_calls)
 
+  @mock.patch('google_compute_engine.accounts.accounts_utils.file_utils.open')
+  @mock.patch('google_compute_engine.accounts.accounts_utils.os.path.exists')
+  def testCreateSudoersGroupWriteError(
+      self, mock_exists, mock_open):
```
illfelder: This should be able to fit on one line.

google_compute_engine/accounts/accounts_utils.py
```diff
@@ -76,10 +76,16 @@ def _CreateSudoersGroup(self):
...
+      try:
+        with open(self.google_sudoers_file, 'w') as group:
+          message = '%{0} ALL=(ALL:ALL) NOPASSWD:ALL'.format(
+              self.google_sudoers_group)
+          group.write(message)
+      except IOError as e:
+        self.logger.error(
+            'Could not write sudoers file. %s\n%s' % (
+                self.google_sudoers_file, str(e)))
+        return None
```
illfelder: This function doesn't have a return type. Instead, you can add an `else:` that will then set permissions on the file.

google_compute_engine/accounts/accounts_utils.py
```diff
@@ -76,10 +76,16 @@ def _CreateSudoersGroup(self):
...
-            self.google_sudoers_group)
-        group.write(message)
+      try:
+        with open(self.google_sudoers_file, 'w') as group:
+          message = '%{0} ALL=(ALL:ALL) NOPASSWD:ALL'.format(
+              self.google_sudoers_group)
+          group.write(message)
+      except IOError as e:
+        self.logger.error(
+            'Could not write sudoers file. %s\n%s' % (
```
illfelder: `'Could not write sudoers file: %s. %s.'`

google_compute_engine/accounts/tests/accounts_utils_test.py
```diff
@@ -148,6 +149,23 @@ def testCreateSudoersGroupError(
...
+    mocks = mock.Mock()
```
illfelder: You're not actually using this in your test.

google_compute_engine/accounts/tests/accounts_utils_test.py
```diff
@@ -148,6 +149,23 @@ def testCreateSudoersGroupError(
...
+    mocks.attach_mock(mock_exists, 'exists')
+    mocks.attach_mock(mock_open, 'open')
+    mocks.attach_mock(self.mock_utils._GetGroup, 'group')
+    mocks.attach_mock(self.mock_logger, 'logger')
+    self.mock_utils._GetGroup.return_value = True
+    mock_exists.return_value = False
+    mock_open.side_effect = IOError
+
+    self.assertIsNone(
```
illfelder: You should check that the correct functions are called when the file write fails. The check should be similar to testCreateSudoersGroupError.

google_compute_engine/accounts/tests/accounts_utils_test.py
```diff
@@ -148,6 +149,23 @@ def testCreateSudoersGroupError(
     ]
     self.assertEqual(mocks.mock_calls, expected_calls)
 
+  @mock.patch('google_compute_engine.accounts.accounts_utils.file_utils.open')
+  @mock.patch('google_compute_engine.accounts.accounts_utils.os.path.exists')
+  def testCreateSudoersGroupWriteError(
```
illfelder: Currently this test is failing on Python 3.3 and 3.4.

google_compute_engine/accounts/accounts_utils.py
rjschwei: Setting the permissions in an else: clause changes the logic of the implementation. Previously the permissions on the file were set as desired, whether it was written by this code or already in existence, moving the code to an else: implies that permissions are only set when the file is written by this code.

I suppose it is fair to assume that the file has the proper permissions if it exists.

illfelder: Use `return` instead of `return None` (they are equivalent, but return None is redundant).

google_compute_engine/accounts/tests/accounts_utils_test.py
illfelder: To clarify why this should be changed, the functional always returns None so this isn't checking anything.

google_compute_engine/accounts/accounts_utils.py
rjschwei: I know, in other parts of the code "return None" is used, was trying to be consistent.

google_compute_engine/accounts/tests/accounts_utils_test.py
rjschwei: Yup, makes sense.

rjschwei:
OK, lets see how this goes now

illfelder:
Can you clean up your github branch so that it only shows the relevant commits?

rjschwei:
Again trying to push to master, sorry. Habbits.

PR #352, opened 2016-10-19
- Fix up the dependencies

rjschwei:
- The setup.py file is used to generate scripts. The generated scripts
  create a runtime dependence on the setuptools, thus these must be
  installed on the target system where the code is supposed to run.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,7 +25,7 @@ setuptools.setup(
     author_email='gc-team@google.com',
     description='Google Compute Engine',
     include_package_data=True,
-    install_requires=['boto'],
+    install_requires=['boto', 'setuptools'],
     license='Apache Software License',
     long_description='Google Compute Engine guest environment.',
     name='google-compute-engine',


illfelder:
I confirmed this is needed. Thanks for the contribution!

helen-fornazier:
Hi,

Just for clarification, if I understand correct, install_requirements is used for run time dependencies, and it seems that none of the scripts uses setuptools after the package is installed. or is there any usage that I am missing here?

I am asking this because I notice that in FreeBSD they make the following relation between their packaging variables and setuptools:
* setup_requires -> BUILD_DEPENDS
* install_requires -> RUN_DEPENDS
and I don't see why the FreeBSD package should add setuptools in the RUN_DEPENDS instead of BUILD_DEPENDS
so I was wondering if we could remove the setuptools from the install_requires

Thanks

zmarano:
The entry point scripts that get generated from setuptools uses it... /usr/bin/google_instance_setup ... etc. Its both a build dependency and run dependency.

PR #375, opened 2017-02-24
Support >32 vCPU's in set_multiqueue.

zmarano:
Have the package conflict with irqbalance which overwrites these settings.

### DIFF
--- a/scripts/set_multiqueue
+++ b/scripts/set_multiqueue
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright 2016 Google Inc. All Rights Reserved.
+# Copyright 2017 Google Inc. All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


scripts/set_multiqueue
```diff
@@ -35,7 +35,7 @@ function set_channels() {
   ethtool -L "${1}" combined "${2}" > /dev/null 2>&1
 }
 
-echo "Running $(basename "$0")."
+echo "Running $(basename $0)"
```
illfelder: Please add the period back to the end.

scripts/set_multiqueue
```diff
@@ -78,56 +78,58 @@ do
       # a decimal integer ranging from 0 to K - 1 where K is the number of
       # input (output) queues in the virtionet device.
       virtionet_intx_dir="${irq}/${dev}"
-      virtionet_msix_dir_regex=".*/${dev}-(input|output)\.[0-9]+$"
+      virtionet_msix_dir_regex=".*/${dev}-(input|output)\.([0-9]+)$"
       if [ -d "${virtionet_intx_dir}" ]; then
         # All virtionet intx IRQs are delivered to CPU 0
         echo "Setting ${smp_affinity} to 01 for device ${dev}."
-        echo "01" > "${smp_affinity}"
+        echo "01" > ${smp_affinity}
```
illfelder: You shouldn't need to remove this quoting.

PR #378, opened 2017-03-02
Remove irqbalance conflicts which doesn't correctly update on RPM systems.

zmarano:


setup.py
```diff
@@ -32,7 +32,7 @@
     packages=setuptools.find_packages(),
     scripts=glob.glob('scripts/*'),
     url='https://github.com/GoogleCloudPlatform/compute-image-packages',
-    version='2.3.3',
+    version='2.3.4',
```
illfelder: Move this to a separate commit. You can also bump the other package in that commit.

PR #385, opened 2017-03-13
home dir SetPermissions changed from 755 to 750

ryancurrah:
Sorry for not branching from development in the first place. I have also fixed the failing test, I think.


illfelder:
LGMT. Thanks for making this change!

PR #409, opened 2017-06-07
Update instance setup to accept metadata configs.

illfelder:
If specified, instance metadata or project metadata will take precedence
over configs specified in the `/etc/default/` files.

### DIFF
--- a/google_compute_engine/instance_setup/instance_setup.py
+++ b/google_compute_engine/instance_setup/instance_setup.py
@@ -48,21 +48,43 @@ class InstanceSetup(object):
     self.metadata_dict = None
     self.instance_config = instance_config.InstanceConfig()
 
+    if self.instance_config.GetOptionBool('InstanceSetup', 'network_enabled'):
+      self.metadata_dict = self.watcher.GetMetadata()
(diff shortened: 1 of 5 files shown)


google_compute_engine/instance_setup/instance_setup.py
```diff
@@ -48,22 +48,44 @@ def __init__(self, debug=False):
...
+      instance_data = {}
+      self.logger.warning('Instance attributes were not found.')
+
+    try:
+      project_data = self.metadata_dict['project']['attributes']
+    except KeyError:
+      project_data = {}
+      self.logger.warning('Project attributes were not found.')
+
+    return (instance_data.get('instance-configs') or
```
adityakali: Do you think "instance-configs" may be too generic name for a key?
What happens if the data in this key is not as per the format expected by this script?

illfelder: Great feedback. I've made a new pull request:
https://github.com/GoogleCloudPlatform/compute-image-packages/pull/411

I've changed the name to `google-instance-configs` and I've added error handling for both the metadata value, and for the config files. If the file cannot be parsed, it won't be used.

PR #413, opened 2017-06-14
Add stricter type checking for JSON parse methods.

Sarsate:
Some json-c methods, such as json_object_object_foreach, will segfault
if they are given a given a json_object of a different type than expected. 
This meant that a malformed response from the metadata server
could cause a segfault in the guest.

This adds stricter type checking around these calls to cleanly handle
malformed responses.

### DIFF
--- a/google_compute_engine_oslogin/utils/oslogin_utils.cc
+++ b/google_compute_engine_oslogin/utils/oslogin_utils.cc
@@ -114,6 +114,9 @@ bool NssCache::LoadJsonArrayToCache(string response) {
     page_token_ = "";
     return false;
   }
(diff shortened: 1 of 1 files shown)


google_compute_engine_oslogin/utils/oslogin_utils.cc
```diff
@@ -234,19 +240,32 @@ std::vector<string> ParseJsonToSshKeys(string response) {
     return result;
   }
 
+  if(json_object_get_type(ssh_public_keys) != json_type_object) {
```
illfelder: space in between the if statement and the paren.

Sarsate: Done.

Issue #424, opened 2017-06-23
Update OS Login sudo file permissions to restrict read access.


illfelder:
A non-sudo user shouldn't be able to see every user with sudo access in the VM.

Issue #423, opened 2017-06-23
OS Login sudoer files should be written to a var directory.


illfelder:
Currently we write a file for each sudoer in `/etc/google_sudoers/`. It would make more sense if this location was `/var/google-sudoers.d/`.

PR #425, opened 2017-06-23
Update sudoers directory location and permission.

illfelder:
As part of this update, we will also (likely) need to update the SELinux
context for EL distros.

#423 #424


illfelder:
We've confirmed that changes to the SELinux configuration are needed for this change. We'll make a followup pull request with the new `te` and `pp` files.

PR #431, opened 2017-07-05
Retry HTTP requests if error 500 is received.

Sarsate:
HTTP response code 500 is likely a transient error. This is worth
retrying for to make the guest code a bit more resilient.

The retry count is currently set to 1 to reduce the risk of blocking
various nss methods in the event of a persistent backend failure.


illfelder:
Thanks!

PR #432, opened 2017-07-06
Add check for SLES to restart sshd.

zmarano:


google_compute_engine_oslogin/bin/google_oslogin_control
```diff
@@ -98,8 +98,10 @@ restart_sshd() {
     echo "Restarting sshd."
     if which service >/dev/null 2>&1
     then
-      if [ -e ${el_release_file} ]; then
+      # EL and SLES use sshd.
+      if [ -e ${el_release_file} -o -d /etc/zypp]; then
```
illfelder: How about:
```
if service --status-all | grep -Fq 'sshd'; then    
  service sshd restart
fi
```

zmarano: Yeah ok that's slightly better.

PR #435, opened 2017-07-12
Permit public https gs files as normal download

ivanseidel:
COS (Container-Optimized OS from Google) doesn't comes pre-installed with `gsutil`.

If you try to setup `startup-script-url` metadata using Google Storage public link url (eg: `https://storage.googleapis/<bucket>/<object>`), it will fail: 

```terminal
myuser@some-instance-d8bw ~ $ sudo google_metadata_script_runner -d --script-type=startup
startup-script: INFO Starting startup scripts.
startup-script: INFO Found startup-script-url in metadata.
startup-script: WARNING gsutil is not installed, cannot download items from Google Storage.
startup-script: INFO No startup scripts found in metadata.
startup-script: INFO Finished running startup scripts.
```

The thing is: I had to dig a bit deeper to find this. But, my startup script is publicly available on Google Storage, and it only failed because it didn't attempt with usual `_DownloadUrl` fetching.

This patch allows for all users from Google Container optimized-OS to use `startup-script-url` stored on google storage.

### DIFF
--- a/google_compute_engine/metadata_scripts/script_retriever.py
+++ b/google_compute_engine/metadata_scripts/script_retriever.py
@@ -135,10 +135,8 @@ class ScriptRetriever(object):
     match = gs_regex.match(url)
     if match:
       gs_url = r'gs://%s/%s' % (match.group('bucket'), match.group('obj'))
-      dest = self._DownloadGsUrl(gs_url, dest_dir)
       # In case gsutil is not installed, continue as a normal URL
-      if dest is not None:
-        return dest
+      return self._DownloadGsUrl(gs_url, dest_dir) or self._DownloadUrl(url, dest_dir)
 
     # Check for the other possible Google Storage URLs:
     # http://storage.googleapis.com/<bucket>/<object>
@@ -152,10 +150,8 @@ class ScriptRetriever(object):
     match = gs_regex.match(url)
     if match:
       gs_url = r'gs://%s/%s' % (match.group('bucket'), match.group('obj'))
-      dest = self._DownloadGsUrl(gs_url, dest_dir)
       # In case gsutil is not installed, continue as a normal URL
-      if dest is not None:
-        return dest
+      return self._DownloadGsUrl(gs_url, dest_dir) or self._DownloadUrl(url, dest_dir)
 
     # Unauthenticated download of the object.
     return self._DownloadUrl(url, dest_dir)
--- a/google_compute_engine/metadata_scripts/script_retriever.py
+++ b/google_compute_engine/metadata_scripts/script_retriever.py
@@ -135,7 +135,7 @@ class ScriptRetriever(object):
     match = gs_regex.match(url)
     if match:
       gs_url = r'gs://%s/%s' % (match.group('bucket'), match.group('obj'))
-      # In case gsutil is not installed, continue as a normal URL
+      # In case gsutil is not installed, continue as a normal URL.
       return (self._DownloadGsUrl(gs_url, dest_dir) or
               self._DownloadUrl(url, dest_dir))
 
@@ -151,7 +151,7 @@ class ScriptRetriever(object):
     match = gs_regex.match(url)
     if match:
       gs_url = r'gs://%s/%s' % (match.group('bucket'), match.group('obj'))
-      # In case gsutil is not installed, continue as a normal URL
+      # In case gsutil is not installed, continue as a normal URL.
       return (self._DownloadGsUrl(gs_url, dest_dir) or
               self._DownloadUrl(url, dest_dir))
 
--- a/google_compute_engine/metadata_scripts/script_retriever.py
+++ b/google_compute_engine/metadata_scripts/script_retriever.py
@@ -135,7 +135,10 @@ class ScriptRetriever(object):
     match = gs_regex.match(url)
     if match:
       gs_url = r'gs://%s/%s' % (match.group('bucket'), match.group('obj'))
-      return self._DownloadGsUrl(gs_url, dest_dir)
+      dest = self._DownloadGsUrl(gs_url, dest_dir)
+      # In case gsutil is not installed, continue as a normal URL
+      if dest is not None:
+        return dest
 
     # Check for the other possible Google Storage URLs:
     # http://storage.googleapis.com/<bucket>/<object>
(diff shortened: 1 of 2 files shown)


wonderfly:
Patch seems fine, though I want to point out that this will only allow COS users to use publicly accessible `startup-script-url`s, not all of them.

google_compute_engine/metadata_scripts/script_retriever.py
```diff
@@ -135,7 +135,10 @@ def _DownloadScript(self, url, dest_dir):
     match = gs_regex.match(url)
     if match:
       gs_url = r'gs://%s/%s' % (match.group('bucket'), match.group('obj'))
-      return self._DownloadGsUrl(gs_url, dest_dir)
+      dest = self._DownloadGsUrl(gs_url, dest_dir)
+      # In case gsutil is not installed, continue as a normal URL
+      if dest is not None:
+        return dest
```
illfelder: Instead of doing this, define a variable at the top of the file called `dest_file`, in each call to `_DownloadGsUrl` save the output to the variable, and then `return dest_file or self._DownloadUrl(url, dest_dir)`.

Also please add testing for this.

ivanseidel: Or should we call `_DownloadUrl` from within `_DownloadGsUrl` if it matches `https://` or `http://`?

illfelder: `self._DownloadGsUrl` is always called with a url of the form `gs://<bucket>/<object>` so that wouldn't help.

ivanseidel: It's possible to be just:
```
# In case gsutil is not installed, continue as a normal URL
return self._DownloadGsUrl(gs_url, dest_dir) or self._DownloadUrl(url, dest_dir)
```
Go/NoGo?

illfelder: That would be for every case where we try to use the Google Storage URL function. I think it makes the most sense to do it once in the function return as a fallback.

ivanseidel: Right not there are only two cases. The other one wouldn't work:
```
if url.startswith(r'gs://'):
      return self._DownloadGsUrl(url, dest_dir)
```

illfelder: Sure. It's fine to do the disjunction on the two cases then.

ivanseidel:
It should be fine now... (Tested locally removing modifications from `script_retriever.py` and it failed.

Please check if I'm asserting for everything.

ivanseidel:
Not sure why it failed, but only happened to python `3.x`. Any ideas?

google_compute_engine/metadata_scripts/script_retriever.py
```diff
@@ -135,7 +135,8 @@ def _DownloadScript(self, url, dest_dir):
     match = gs_regex.match(url)
     if match:
       gs_url = r'gs://%s/%s' % (match.group('bucket'), match.group('obj'))
-      return self._DownloadGsUrl(gs_url, dest_dir)
+      # In case gsutil is not installed, continue as a normal URL
+      return self._DownloadGsUrl(gs_url, dest_dir) or self._DownloadUrl(url, dest_dir)
```
illfelder: This is over 80 characters. Please change this to:

```
return (self._DownloadGsUrl(gs_url, dest_dir) or
        self._DownloadUrl(url, dest_dir))

File: google_compute_engine/metadata_scripts/script_retriever.py [hunk 3]
```diff
@@ -149,7 +150,8 @@ def _DownloadScript(self, url, dest_dir):
...
```
illfelder: Here as well.

google_compute_engine/metadata_scripts/tests/script_retriever_test.py
```diff
@@ -44,6 +44,21 @@ def testDownloadGsNoExec(self, mock_call):
 
   @mock.patch('google_compute_engine.metadata_scripts.script_retriever.subprocess.check_call')
   @mock.patch('google_compute_engine.metadata_scripts.script_retriever.tempfile.NamedTemporaryFile')
+  def testDownloadGsUrlFallback(self, mock_tempfile, mock_call):
```
illfelder: This test should be combined with `testDownloadScript` so that you can test each url type. Currently you have an issue where you are relying on `_DownloadUrl` returning the correct value, but that isn't the function this test is testing, so the mocks probably aren't set up correctly in Python 3.

ivanseidel: Any ideas on how to prevent `gsutil` from running during the test and causing it to fetch other endpoint?
I think we can basically check if disabling `gsutil` still returns the original URL on a few cases

illfelder:
Created #436 to include your changes with testing.

wonderfly:
> Created #436 to include your changes with testing.

I think as a maintainer of the project you get to modify his commit inline so he gets credit too. :)

google_compute_engine/metadata_scripts/tests/script_retriever_test.py
illfelder: I don't think that was the issue here. I think the test was failing because:

1. `self._DownloadGsUrl(gs_url, dest_dir)` threw an error, returning `None`.
2. `self._DownloadUrl(url, dest_dir)` was called with the original URL, threw an error, and the returned `None` instead of `self.dest_dir`.

The reason I suggested combining this test with `testDownloadScript` is to isolate the logic you changed from the rest of functions we have already validated are working properly. This test is currently checking the combined behavior of several functions, which is more of an integration test.

ivanseidel:
Why did you copy my commits @illfelder?

Please submit a solution for testing and let's keep this issue as the main one. I appreciate if you use this instead of creating another one with the same purpose.

I need help with testing. Not sure why Python 3.x is failing.

illfelder:
I commented about the testing issues inline, but to answer your question, I forked the commit that downloads using curl if the google storage download fails and created a separate PR in order to implement the testing changes. If you like, you can revert the testing file, fix a couple issues in `script_retriever.py` and we can check in the changes separately.

google_compute_engine/metadata_scripts/script_retriever.py
```diff
@@ -135,7 +135,8 @@ def _DownloadScript(self, url, dest_dir):
     match = gs_regex.match(url)
     if match:
       gs_url = r'gs://%s/%s' % (match.group('bucket'), match.group('obj'))
-      return self._DownloadGsUrl(gs_url, dest_dir)
+      # In case gsutil is not installed, continue as a normal URL
```
illfelder: Please end the comment with a period.

google_compute_engine/metadata_scripts/script_retriever.py
```diff
@@ -149,7 +150,8 @@ def _DownloadScript(self, url, dest_dir):
...
```
illfelder: And here.

ivanseidel:
Awesome, thanks! I will update it and let you know

illfelder:
We are preparing to make a new GitHub release, so I'd like to merge in this change. I'll wait a couple hours in case you still want to update this PR, but otherwise I'll merge in #436 and close this off.

PR #440, opened 2017-07-19
Replace file paths by variables

helen-fornazier:
Paths can vary depending on the system. Instead of using the full path
in the code, use variables containing the correct path according to the
system. This will decrease the amount of patches required for FreeBSD,
OpenBSD, etc. and also will make the package easier to maintain across
different systems.

### DIFF
--- a/google_compute_engine/accounts/accounts_daemon.py
+++ b/google_compute_engine/accounts/accounts_daemon.py
@@ -22,12 +22,13 @@ import optparse
 import random
 
 from google_compute_engine import config_manager
+from google_compute_engine import constants
 from google_compute_engine import file_utils
 from google_compute_engine import logger
 from google_compute_engine import metadata_watcher
 from google_compute_engine.accounts import accounts_utils
 
-LOCKFILE = '/var/lock/google_accounts.lock'
+LOCKFILE = constants.LOCALSTATEDIR + '/lock/google_accounts.lock'
 
 
 class AccountsDaemon(object):
--- a/google_compute_engine/accounts/accounts_utils.py
+++ b/google_compute_engine/accounts/accounts_utils.py
@@ -23,6 +23,7 @@ import shutil
 import subprocess
 import tempfile
 
+from google_compute_engine import constants
 from google_compute_engine import file_utils
 
 USER_REGEX = re.compile(r'\A[A-Za-z0-9._][A-Za-z0-9._-]*\Z')
@@ -43,8 +44,9 @@ class AccountsUtils(object):
     """
     self.logger = logger
     self.google_sudoers_group = 'google-sudoers'
-    self.google_sudoers_file = '/etc/sudoers.d/google_sudoers'
-    self.google_users_dir = '/var/lib/google'
+    self.google_sudoers_file = (
+        constants.LOCALBASE + '/etc/sudoers.d/google_sudoers')
+    self.google_users_dir = constants.LOCALBASE + '/var/lib/google'
     self.google_users_file = os.path.join(self.google_users_dir, 'google_users')
 
     self._CreateSudoersGroup()
--- a/google_compute_engine/config_manager.py
+++ b/google_compute_engine/config_manager.py
@@ -18,10 +18,11 @@
 import os
 import textwrap
 
+from google_compute_engine import constants
 from google_compute_engine import file_utils
 from google_compute_engine.compat import parser
 
-CONFIG = '/etc/default/instance_configs.cfg'
+CONFIG = constants.SYSCONFDIR + '/instance_configs.cfg'
 
 
 class ConfigManager(object):
@@ -101,7 +102,8 @@ class ConfigManager(object):
     """
     config_file = config_file or self.config_file
     config_name = os.path.splitext(os.path.basename(config_file))[0]
-    config_lock = '/var/lock/google_%s.lock' % config_name
+    config_lock = (
+        '%s/lock/google_%s.lock' % (constants.LOCALSTATEDIR, config_name))
     with file_utils.LockFile(config_lock):
       with open(config_file, 'w') as config_fp:
         if self.config_header:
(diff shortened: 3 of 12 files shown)


illfelder:
In general, it's better practice to import an entire module rather than the particular functions/variables you want to use. To support distro specific constants, I think it would make sense to create a `google_compute_engine/constants.py`that you would import in each of the files where you need to use a variable.

google_compute_engine/accounts/accounts_daemon.py
```diff
@@ -23,11 +23,12 @@
 
 from google_compute_engine import config_manager
 from google_compute_engine import file_utils
+from google_compute_engine import LOCALSTATEDIR
 from google_compute_engine import logger
 from google_compute_engine import metadata_watcher
 from google_compute_engine.accounts import accounts_utils
 
-LOCKFILE = '/var/lock/google_accounts.lock'
+LOCKFILE = LOCALSTATEDIR+'/lock/google_accounts.lock'
```
illfelder: Please add a space around both sides of the `+`. (Apply this comment everywhere).

helen-fornazier: Thanks for your review. I just updated this and I moved the constants to contants.py as you suggested. Please, let me know if there is anything else I could improve.

google_compute_engine/config_manager.py
```diff
@@ -101,7 +102,7 @@ def WriteConfig(self, config_file=None):
     """
     config_file = config_file or self.config_file
     config_name = os.path.splitext(os.path.basename(config_file))[0]
-    config_lock = '/var/lock/google_%s.lock' % config_name
+    config_lock = constants.LOCALSTATEDIR + '/lock/google_%s.lock' % config_name
```
illfelder: In this case where you have a string with multiple variables, its preferable to use:
```
config_lock = ( 
    '%s/lock/google_%s.lock' % (constants.LOCALSTATEDIR, config_name))
```

google_compute_engine/accounts/accounts_utils.py
```diff
@@ -43,8 +44,8 @@ def __init__(self, logger, groups=None, remove=False):
     """
     self.logger = logger
     self.google_sudoers_group = 'google-sudoers'
-    self.google_sudoers_file = '/etc/sudoers.d/google_sudoers'
-    self.google_users_dir = '/var/lib/google'
+    self.google_sudoers_file = constants.LOCALBASE + '/etc/sudoers.d/google_sudoers'
```
illfelder: Please keep line lengths under 80 characters. I'd suggest updating this with:

```
self.google_sudoers_file = (
    constants.LOCALBASE + '/etc/sudoers.d/google_sudoers')
````

File: google_compute_engine/constants.py [hunk 4]
```diff
@@ -0,0 +1,36 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""A module for global constants."""
+
+import platform
+
+if platform.system() == "FreeBSD":
```
illfelder: Please replace all double quotes with single quotes.

helen-fornazier:
@illfelder Thanks again for your review. I just sent an update with your suggested changes, I hope I caught all the coding style issues this time.

google_compute_engine/constants.py
```diff
@@ -0,0 +1,36 @@
...
+"""A module for global constants."""
+
+import platform
+
+if platform.system() == 'FreeBSD':
+    LOCALBASE = '/usr/local'
+    BOTOCONFDIR = '/usr/local'
+    SYSCONFDIR = '/usr/local/etc'
+    LOCALSTATEDIR = '/var/spool'
+
```
illfelder: Please remove whitespace between the if, elseif, else

google_compute_engine/constants.py
```diff
@@ -0,0 +1,36 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""A module for global constants."""
+
+import platform
+
+if platform.system() == 'FreeBSD':
+    LOCALBASE = '/usr/local'
```
illfelder: Overall this change looks reasonable. One last request - would you mind separating the words in the variable names with underscores to make it easier to read?

google_compute_engine/network_setup/network_setup.py
```diff
@@ -130,7 +131,7 @@ def _EnableNetworkInterfaces(self, interfaces):
       except subprocess.CalledProcessError:
         self.logger.warning('Could not enable Ethernet interfaces.')
     else:
-      if os.path.exists('/etc/sysconfig/network-scripts'):
+      if os.path.exists(constants.LOCALBASE + '/etc/sysconfig/network-scripts'):
```
illfelder: If you rename the variable to `LOCAL_BASE`, this will be over 80 characters. I'd suggest creating a class variable (below network_interfaces) called `network_path` and then use `self.network_path` here and in `_DisableNetworkManager`. Given that both use these string constants, it's better to do that anyway.

google_compute_engine/constants.py
helen-fornazier: I was thinking to leave them like this for similarity to other systems, for example https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html
But if you still prefer I can add the underscores, no problem, then it would be:
```
    LOCAL_BASE
    BOTO_CONF_DIR
    SYS_CONF_DIR
    LOCAL_STATE_DIR

```

illfelder: Thanks for the explanation. Feel free to leave it as is.

google_compute_engine/constants.py
illfelder: After you update this I'll approve and merge.

helen-fornazier:
@illfelder I added the network_path (event if LOCALBASE is not renamed), so there is a single attribution. I had to change a line in the mocks but I think it is ok, what do you think?

PR #442, opened 2017-07-21
Use netifaces for getting mac address

helen-fornazier:
BSD doesn't have the path /sys/class/net/<interface>/address
Use netifaces instead for interoperability with BSDs systems

### DIFF
--- a/google_compute_engine/network_utils.py
+++ b/google_compute_engine/network_utils.py
@@ -17,6 +17,14 @@
 
 import logging
 import os
+import re
+try:
+  import netifaces
+except ImportError:
+  netifaces = None
+
+
+MAC_REGEX = re.compile(r'\A([0-9A-Fa-f]{2}[:]){5}([0-9A-Fa-f]{2})\Z')
 
 
 class NetworkUtils(object):
@@ -37,6 +45,17 @@ class NetworkUtils(object):
     Returns:
       dict, string MAC addresses mapped to the string network interface name.
     """
+    if netifaces:
+      return self._CreateInterfaceMapNetifaces()
+    else:
+      return self._CreateInterfaceMapSysfs()
+
+  def _CreateInterfaceMapSysfs(self):
+    """Generate a dictionary mapping MAC address to Ethernet interfaces.
+
+    Returns:
+      dict, string MAC addresses mapped to the string network interface name.
+    """
     interfaces = {}
     for interface in os.listdir('/sys/class/net'):
       try:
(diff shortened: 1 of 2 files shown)


google_compute_engine/network_utils.py
```diff
@@ -38,11 +44,16 @@ def _CreateInterfaceMap(self):
...
     """
     interfaces = {}
-    for interface in os.listdir('/sys/class/net'):
+    for interface in netifaces.interfaces():
       try:
-        mac_address = open(
-            '/sys/class/net/%s/address' % interface).read().strip()
-      except (IOError, OSError) as e:
+        mac_address = (
+            netifaces.ifaddresses(interface)[netifaces.AF_LINK][0]['addr'])
```
illfelder: A more safe way to do this:

```
af_link = netifaces.ifaddresses(interface).get(netifaces.AF_LINK, [])
mac_address = next(iter(af_link), {}).get('addr')
```

google_compute_engine/network_utils.py
```diff
@@ -38,11 +44,16 @@ def _CreateInterfaceMap(self):
...
+        # In some systems this field can come with an empty string or with the
+        # name of the interface when there is no MAC address associated with it.
+        # Check the regex to be sure
+        if not MAC_REGEX.match(mac_address):
+            raise Exception('No MAC Address')
+      except Exception as e:
```
illfelder: Can we make this exception any more specific to the possible excepts netifaces might throw?

illfelder:
Thanks for the contribution. When we were writing this logic, we decided against using netifaces because we were apprehensive about the package install on all of the supported distros. That might not be a valid concern, but we will need to spend some time discussing whether `netifaces` is a package we can safely include as a dependency.

google_compute_engine/network_utils.py
helen-fornazier: right, I thought we didn't have to call it in a safe way, as it would though and exception that would be caught. If in the end mac_address is None, the code will raise an exception anyway. But I don't mind changing if you prefer like this.

illfelder: Yeah, understand. Exceptions in Python are slow so it's good to avoid them except for in exceptional cases (or when it's very inconvenient).

helen-fornazier:
I understand. Thank you for reviewing this anyway.

google_compute_engine/network_utils.py
helen-fornazier: I see. I changed the approach then, it shouldn't be any exceptions raised by netifaces in normal cases (only in extreme cases as a lack of memory for example), so I used the code you suggested and removed the try/catch statement.

helen-fornazier:
Hi,

I tested all the supported distros listed in the README. All of them provides a python-netifaces package, and in almost all of them I could also test using the pip instalation (only SLES11 that doesn't provide a pip package and it needs to be installed from source, but it provides a python-netifaces available through zypper)

To test the netifaces solution, I added both NetworkUtils classes (with netifaces and the current one without it) in this file for testing, I call _CreateInterfaceMap() in both of them and compare the results.
The testing script can be found at https://people.collabora.com/~koike/netiface-test.py

Here is a more detailed report of netifaces in each distro: https://people.collabora.com/~koike/notes-netifaces.txt

I hope this helps deciding if adopting netifaces is ok.

illfelder:
Hi Helen,

I've been thinking of how to support BSD without adding a dependency on all distros. You should be able to do something like this:

```
import pkgutil

if pkgutil.find_loader('netifaces'):
  <your function>
else:
  <existing function>
```

Let me know what you think. If we go down this path, please add basic test coverage for the function.

illfelder:
Hi Helen, are you still interested in making the change?

helen-fornazier:
Right, we can do that and add a requirement in the BSD package adding netifaces as dependency. I'll update this and send you a new version

helen-fornazier:
@illfelder Hi, could you check this last version I uploaded when you have time please?
Thanks

google_compute_engine/network_utils.py
```diff
@@ -17,6 +17,14 @@
 
 import logging
 import os
+import re
+try:
+  import netifaces
+except ImportError:
+  netifaces = False
```
illfelder: s/False/None

illfelder:
Overall change looks good. Sorry for the delay.

PR #443, opened 2017-07-21
Add instance config options for user/group commands.

helen-fornazier:
Command to manage users and groups can vary according to the system.
Thus move this commands to the port.py for interoperability with BSD

The content of constants.py was moved to port.py, as it won't contain only constants variables, the file name is no more appropriated

### DIFF
--- a/google_compute_engine/accounts/accounts_daemon.py
+++ b/google_compute_engine/accounts/accounts_daemon.py
@@ -37,12 +37,18 @@ class AccountsDaemon(object):
   invalid_users = set()
   user_ssh_keys = {}
 
-  def __init__(self, groups=None, remove=False, debug=False):
+  def __init__(
+      self, groups=None, remove=False, useradd_cmd=None, userdel_cmd=None,
+      usermod_cmd=None, groupadd_cmd=None, debug=False):
     """Constructor.
 
     Args:
       groups: string, a comma separated list of groups.
       remove: bool, True if deprovisioning a user should be destructive.
+      useradd_cmd: string, command to create a new user.
+      userdel_cmd: string, command to delete a user.
+      usermod_cmd: string, command to modify user's groups.
+      groupadd_cmd: string, command to add a new group.
       debug: bool, True if debug output should write to the console.
     """
     facility = logging.handlers.SysLogHandler.LOG_DAEMON
@@ -50,7 +56,9 @@ class AccountsDaemon(object):
         name='google-accounts', debug=debug, facility=facility)
     self.watcher = metadata_watcher.MetadataWatcher(logger=self.logger)
     self.utils = accounts_utils.AccountsUtils(
-        logger=self.logger, groups=groups, remove=remove)
+        logger=self.logger, groups=groups, remove=remove,
+        useradd_cmd=useradd_cmd, userdel_cmd=userdel_cmd,
+        usermod_cmd=usermod_cmd, groupadd_cmd=groupadd_cmd)
     try:
       with file_utils.LockFile(LOCKFILE):
         self.logger.info('Starting Google Accounts daemon.')
@@ -226,6 +234,10 @@ def main():
     AccountsDaemon(
         groups=instance_config.GetOptionString('Accounts', 'groups'),
         remove=instance_config.GetOptionBool('Accounts', 'deprovision_remove'),
+        useradd_cmd=instance_config.GetOptionBool('Accounts', 'useradd_cmd'),
+        userdel_cmd=instance_config.GetOptionBool('Accounts', 'userdel_cmd'),
+        usermod_cmd=instance_config.GetOptionBool('Accounts', 'usermod_cmd'),
+        groupadd_cmd=instance_config.GetOptionBool('Accounts', 'groupadd_cmd'),
         debug=bool(options.debug))
 
 
--- a/google_compute_engine/accounts/accounts_utils.py
+++ b/google_compute_engine/accounts/accounts_utils.py
@@ -27,6 +27,10 @@ from google_compute_engine import constants
 from google_compute_engine import file_utils
 
 USER_REGEX = re.compile(r'\A[A-Za-z0-9._][A-Za-z0-9._-]*\Z')
+DEFAULT_USERADD_CMD = 'useradd -m -s /bin/bash -p * {user}'
+DEFAULT_USERDEL_CMD = 'userdel -r {user}'
+DEFAULT_USERMOD_CMD = 'groupadd {group}'
+DEFAULT_GROUPADD_CMD = 'groupadd {group}'
 
 
 class AccountsUtils(object):
@@ -34,14 +38,24 @@ class AccountsUtils(object):
 
   google_comment = '# Added by Google'
 
-  def __init__(self, logger, groups=None, remove=False):
+  def __init__(
+      self, logger, groups=None, remove=False, useradd_cmd=None,
+      userdel_cmd=None, usermod_cmd=None, groupadd_cmd=None):
     """Constructor.
 
     Args:
       logger: logger object, used to write to SysLog and serial port.
       groups: string, a comma separated list of groups.
       remove: bool, True if deprovisioning a user should be destructive.
+      useradd_cmd: string, command to create a new user.
+      userdel_cmd: string, command to delete a user.
+      usermod_cmd: string, command to modify user's groups.
+      groupadd_cmd: string, command to add a new group.
     """
+    self.useradd_cmd = useradd_cmd or DEFAULT_USERADD_CMD
+    self.userdel_cmd = userdel_cmd or DEFAULT_USERDEL_CMD
+    self.usermod_cmd = usermod_cmd or DEFAULT_USERMOD_CMD
+    self.groupadd_cmd = groupadd_cmd or DEFAULT_GROUPADD_CMD
     self.logger = logger
     self.google_sudoers_group = 'google-sudoers'
     self.google_sudoers_file = (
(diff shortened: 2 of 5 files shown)


illfelder:
Thanks for identifying the changes to make the accounts daemon work on BSD. I think the objective of this pull request has merit, but I think these changes are moving us further away from the ideal design of the guest environment (one where distros are able to make the guest work without patching or pushing code upstream). In the existing guest environment, we attempt to do that by providing a config file (and supporting a `.distro` version of it) to configure settings, paths, code to execute.

One idea is that instead of having a `constants.py`(or `ports.py`) file, we can add a `[Constants]` section and specify the four variables we currently know of. To address these new changes, I'm open to suggestions. I'd definitely like a way to allow a distro to call some python function instead of ours.

illfelder:
As a followup note - the easiest thing we could do to support this is add options in the `[Accounts]` section for the functions you want to make configurable. Then, if the distro specifies the option, execute the script they provide with the arguments we are using.

illfelder:
Any update here?

helen-fornazier:
Hi, thanks for your ideas and sorry for my late reply, I'll update this patch as you suggested and re-submit it again.

helen-fornazier:
@illfelder Hi, I updated as you suggested. There are new [Accounts] options and a .distro file per system (but the default which is Linux).
I was adding a [Constants] section to add the paths, but I realized that I need the Paths to be solved before reading the config files, otherwise I we can't know where the config file is.
What do you think?

google_compute_engine/instance_setup/instance_config.py
```diff
@@ -46,6 +46,23 @@ class InstanceConfig(config_manager.ConfigManager):
...
+          # When the token is specified, SSH denies login:
+          # https://github.com/openssh/openssh-portable/blob/master/auth.c
+          #
+          # To solve the issue, make the password '*' which is also recognized
+          # as locked but does not prevent SSH login.
+          'useradd_cmd': 'useradd -m -s /bin/bash -p * {user}',
+          'userdel_cmd': 'userdel -r {user}',
+          'usermod_cmd': 'groupadd {group}',
+          'groupadd_cmd': 'usermod -G {groups} {user}',
+
```
illfelder: Please remove this whitespace.

google_compute_engine/instance_setup/instance_config.py
```diff
@@ -46,6 +46,23 @@ class InstanceConfig(config_manager.ConfigManager):
...
+          #
+          # SSH uses '!' as its locked account token:
+          # https://github.com/openssh/openssh-portable/blob/master/configure.ac
+          #
+          # When the token is specified, SSH denies login:
+          # https://github.com/openssh/openssh-portable/blob/master/auth.c
+          #
+          # To solve the issue, make the password '*' which is also recognized
+          # as locked but does not prevent SSH login.
+          'useradd_cmd': 'useradd -m -s /bin/bash -p * {user}',
```
illfelder: This is a very cool idea. Have you tried it out to see if it works in practice?

helen-fornazier: Yes, I tested in Linux and in BSD

helen-fornazier:
Right, I can remove the .distro files. It would be nice to be able to install the package from `pip` (pypi) or using `setup.py install` also in the BSD systems without the need of further configuration, but editing the configuration file is better then patching in the BSD package at least.

helen-fornazier:
@illfelder, I just resent a new version without adding the distro folder and not touching the setup.py file. What do you think ?

illfelder:
These changes look good to me! Regarding the question about the constants file - I think your comment makes sense and is a good reason to keep it unchanged. If there is a good way to incorporate it into the defaults file, that's still preferred, but we can figure that out at a later time.

google_compute_engine/accounts/accounts_utils.py
```diff
@@ -150,7 +152,7 @@ def _UpdateUserGroups(self, user, groups):
     """
     groups = ','.join(groups)
     self.logger.debug('Updating user %s with groups %s.', user, groups)
-    command = ['usermod', '-G', groups, user]
+    command = self.usermod_cmd.format(user=user, groups=groups)
     try:
       subprocess.check_call(command)
```
illfelder: Given that command is now a string instead of a list, does this check_call still work?

helen-fornazier: check_call accepts a list or a string. A list is usually preferable as it deals with escaping and quoting of arguments, but we are not doing anything complex, I just forget to add shell=True in the call as recommended by the docs

From the docs https://docs.python.org/2/library/subprocess.html:

> args is required for all calls and should be a string, or a sequence of program arguments. Providing a sequence of arguments is generally preferred, as it allows the module to take care of any required escaping and quoting of arguments (e.g. to permit spaces in file names). If passing a single string, either shell must be True (see below) or else the string must simply name the program to be executed without specifying any arguments.

illfelder: Yeah, without adding the shell=True, this will try to execute as a path, and throw a `FileNotFoundError`.

helen-fornazier: Thanks, I think my dev environment is biased as it was working for me like this. Right, sending a new version in a moment.

google_compute_engine/accounts/accounts_utils.py
```diff
@@ -73,7 +85,8 @@ def _CreateSudoersGroup(self):
     """Create a Linux group for Google added sudo user accounts."""
     if not self._GetGroup(self.google_sudoers_group):
       try:
-        subprocess.check_call(['groupadd', self.google_sudoers_group])
+        subprocess.check_call(
+            self.groupadd_cmd.format(group=self.google_sudoers_group), shell=True)
```
illfelder: This is over 80 characters.

helen-fornazier: ops, I am really sorry about that

google_compute_engine/accounts/accounts_utils.py
```diff
@@ -73,7 +85,8 @@ def _CreateSudoersGroup(self):
     """Create a Linux group for Google added sudo user accounts."""
     if not self._GetGroup(self.google_sudoers_group):
       try:
-        subprocess.check_call(['groupadd', self.google_sudoers_group])
+        subprocess.check_call(self.groupadd_cmd.format(
```
illfelder: Would you mind making this:
```
subprocess.check_call(
    self.groupadd_cmd.format(group=self.google_sudoers_group),
    shell=True)
```

google_compute_engine/accounts/accounts_utils.py
illfelder: No worries at all - we need to add an automated linter to this :)

google_compute_engine/accounts/accounts_utils.py
helen-fornazier: done

google_compute_engine/accounts/accounts_daemon.py
```diff
@@ -37,20 +37,30 @@ class AccountsDaemon(object):
   invalid_users = set()
   user_ssh_keys = {}
 
-  def __init__(self, groups=None, remove=False, debug=False):
+  def __init__(
+      self, groups=None, remove=False,
+      useradd_cmd='useradd -m -s /bin/bash -p * {user}',
```
illfelder: One Python style nit - mind pulling these strings out into variables defined at the top of the file?

illfelder: A clarification - can you pull them into variables defined at the top, set them to None here, and then when setting the class variables set them do something like:
```
self.utils = accounts_utils.AccountsUtils(
        logger=self.logger, groups=groups, remove=remove,
        useradd_cmd=useradd_cmd or DEFAULT_USERADD_CMD,
        userdel_cmd=userdel_cmd or DEFAULT_USERDEL_CMD,
        usermod_cmd=usermod_cmd or DEFAULT_USERMOD_CMD,
        groupadd_cmd=groupadd_cmd or DEFAULT_GROUPADD_CMD)
```

helen-fornazier: Done. I updated this in both files, accounts_utils.py and accounts_daemon.py, please check.
Thanks

google_compute_engine/accounts/accounts_daemon.py
```diff
@@ -29,6 +29,10 @@
 from google_compute_engine.accounts import accounts_utils
 
 LOCKFILE = constants.LOCALSTATEDIR + '/lock/google_accounts.lock'
+DEFAULT_USERADD_CMD = 'useradd -m -s /bin/bash -p * {user}'
+DEFAULT_USERDEL_CMD = 'userdel -r {user}'
+DEFAULT_USERMOD_CMD = 'groupadd {group}'
+DEFAULT_GROUPADD_CMD = 'groupadd {group}'
```
illfelder: I missed this in my last review comment - since you only need these variables in the utils class, you can remove their definitions in this file and just have them default to `None`.

google_compute_engine/accounts/accounts_utils.py
```diff
@@ -27,21 +27,36 @@
...
 
 class AccountsUtils(object):
   """System user account configuration utilities."""
 
   google_comment = '# Added by Google'
 
-  def __init__(self, logger, groups=None, remove=False):
+  def __init__(
+      self, logger, groups=None, remove=False, useradd_cmd=DEFAULT_USERADD_CMD,
+      userdel_cmd=DEFAULT_USERDEL_CMD, usermod_cmd=DEFAULT_USERMOD_CMD,
```
illfelder: Please set the variables to None by default, and have the class variables do the `useradd_cmd or DEFAULT_USERADD_CMD` logic when you set them below (lines 56-59).

google_compute_engine/accounts/accounts_daemon.py
helen-fornazier: Done

google_compute_engine/accounts/accounts_utils.py
helen-fornazier: Done. Please check. Thanks

illfelder: Looks great! Thanks!

PR #428, opened 2017-07-03
Reenable testing against pypy3 with pypy3-3.5

rht:
- [x] This is contingent on https://github.com/travis-ci/travis-cookbooks/pull/861.

PR #447, opened 2017-08-02
Reenable testing against pypy3

rht:
A rebase of #428.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: trusty
 language: python
 sudo: true
 python:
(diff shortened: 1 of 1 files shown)


illfelder:
Looks like pypy3 is still failing. Any ideas?

rht:
Still investigating the error; it is due to the pypy3 virtualenv created with tox which somehow misses an env var. https://travis-ci.org/GoogleCloudPlatform/compute-image-packages/jobs/260053408#L520 errs with the requirements file not found.

cclauss:
Would it be useful to resolve the conflicts and then test with PyPy 6.0?
* https://pypy.org/download.html

PR #448, opened 2017-08-02
Modify HttpGet() to return whether the request was successful.

Sarsate:
Changes HttpGet() to return a bool indicating success, and store the response in an out parameter. This lets us differentiate between a backend/server failure, and an empty response.

### DIFF
--- a/google_compute_engine_oslogin/authorized_keys/authorized_keys.cc
+++ b/google_compute_engine_oslogin/authorized_keys/authorized_keys.cc
@@ -36,7 +36,8 @@ int main(int argc, char* argv[]) {
   }
   std::stringstream url;
   url << kMetadataServerUrl << "users?username=" << UrlEncode(argv[1]);
-  string user_response = HttpGet(url.str());
+  string user_response;
+  bool success = HttpGet(url.str(), &response);
   if (user_response.empty()) {
     // Return 0 if the user is not an oslogin user. If we returned a failure
     // code, we would populate auth.log with useless error messages.
@@ -54,7 +55,8 @@ int main(int argc, char* argv[]) {
   url.str("");
   url << kMetadataServerUrl << "authorize?email=" << UrlEncode(email)
       << "&policy=login";
-  string auth_response = HttpGet(url.str());
+  string auth_response;
+  success = HttpGet(url.str(), auth_response);
   if (auth_response.empty()) {
     return 1;
   }
--- a/google_compute_engine_oslogin/authorized_keys/authorized_keys.cc
+++ b/google_compute_engine_oslogin/authorized_keys/authorized_keys.cc
@@ -37,14 +37,16 @@ int main(int argc, char* argv[]) {
   std::stringstream url;
   url << kMetadataServerUrl << "users?username=" << UrlEncode(argv[1]);
   string user_response;
-  bool success = HttpGet(url.str(), &response);
+  if (!HttpGet(url.str(), &user_response)) {
+    return 1;
+  }
   if (user_response.empty()) {
     // Return 0 if the user is not an oslogin user. If we returned a failure
     // code, we would populate auth.log with useless error messages.
     return 0;
   }
   string email = ParseJsonToEmail(user_response);
-  if (email == "") {
+  if (email.empty()) {
     return 1;
   }
   // Redundantly verify that this user has permission to log in to this VM.
@@ -56,8 +58,7 @@ int main(int argc, char* argv[]) {
   url << kMetadataServerUrl << "authorize?email=" << UrlEncode(email)
       << "&policy=login";
   string auth_response;
-  success = HttpGet(url.str(), auth_response);
-  if (auth_response.empty()) {
+  if (!HttpGet(url.str(), &auth_response) || auth_response.empty()) {
     return 1;
   }
   if (!ParseJsonToAuthorizeResponse(auth_response)) {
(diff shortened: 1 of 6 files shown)


wonderfly:
> This lets us differentiate between a backend/server failure, and an empty response.

Is there a way to differentiate between a server failure, and the case where a user cannot be found by the server (ie non osloin user)? With that we can then reliably tell in pam_oslogin_login.cc whether we should return DENIED or SUCCESS.

Sarsate:
If the user can't be found by oslogin, then we should return SUCCESS to avoid preventing a local user from logging in. We could return DENIED on a server failure, but then this would then also block local users in the event of a backend failure. Returning SUCCESS ensures we don't impact a local user's ability to log in, but DENIED would definitely be more secure, so I can see value in both choices. This behavior also only applies to the user lookup--if we fail the authorize call at any point, that will always result in DENIED. 

The risk would be if that exact call failed, the similar lookup in the auth_keys command would also have to fail(and the user also has to have local ssh keys), and all other lookups succeeded, we may accidentally permit a user who shouldn't log in. A persistent backend failure would mean other user lookups would fail, and nss would also not be able to get the user's info, so it is unlikely we would even get this far.

@illfelder any preference on what we should do?

illfelder:
Currently we are discussing a trade-off between security and reliability. In order to prevent blocking local user logins when the OS Login backends are down, and in order to prevent allowing a non-authorized user from logging into the VM, I propose we do the following:

Today, when a user logs in with OS Login, the PAM module will create a file in a particular directory to grant the user sudo permission (assuming the user is authorized). We should create a second directory for non-sudo users. When a non-sudo OS Login user logs in, we can create a file in that second directory. If we get a server error authorizing a user, check if the user has a file associated with them. Deny the user login if they do.

What do you think?

illfelder:
Note: the other benefit of this approach is that we can also address an edge case where a user is deleted from their G Suite organization so the user lookup fails, but the local account still exists.

google_compute_engine_oslogin/pam_module/pam_oslogin_login.cc
```diff
@@ -49,20 +49,20 @@ PAM_EXTERN int pam_sm_acct_mgmt(pam_handle_t *pamh, int flags, int argc,
 
   std::stringstream url;
   url << kMetadataServerUrl << "users?username=" << UrlEncode(str_user_name);
-  string response = HttpGet(url.str());
-  if (response == "") {
+  string response;
+  if (!HttpGet(url.str(), &response) || response.empty()) {
+    // If we are not dealing with an oslogin, the default behavior is to permit.
```
Sarsate: If we return DENIED, then in the event of a persistent backend failure, we block out ALL users, even local ones. No one will be able to log into the VM.

google_compute_engine_oslogin/pam_module/pam_oslogin_admin.cc
```diff
@@ -55,25 +55,24 @@ PAM_EXTERN int pam_sm_acct_mgmt(pam_handle_t *pamh, int flags, int argc,
...
-  if (response == "") {
+  string response;
+  if (!HttpGet(url.str(), &response) || response.empty()) {
     return PAM_SUCCESS;
   }
   string email = ParseJsonToEmail(response);
-  if (email == "") {
+  if (email.empty()) {
     return PAM_SUCCESS;
   }
```
Sarsate: This pam module's control flag is set to optional. The return value has no impact on a user's ability to log in.

google_compute_engine_oslogin/authorized_keys/authorized_keys.cc
```diff
@@ -36,14 +36,17 @@ int main(int argc, char* argv[]) {
   }
   std::stringstream url;
   url << kMetadataServerUrl << "users?username=" << UrlEncode(argv[1]);
-  string user_response = HttpGet(url.str());
+  string user_response;
+  if (!HttpGet(url.str(), &user_response)) {
+    return 1;
+  }
   if (user_response.empty()) {
     // Return 0 if the user is not an oslogin user. If we returned a failure
```
Sarsate: So it actually returns 404, and the response is non-empty, with an error message. This is different from behavior I had tested with a long time ago. It may make sense to return the status code, and have guest code respond to that appropriately.

wonderfly:
Let's lay out the risks of both sides, and see how likely each will happen:
* Solution A. Return DENIED on server failure, the risk is a local user won't be able to login.
* Solution B. Return SUCCESS, the risk is an OSLogin user *with their pubkey stored at /home/\<user\>/.ssh/authorized_keys* will be able to log in.

The risk of Solution A can be realized when there is a *persistent* server failure, and the user info is not already cached by metadata server. Which is highly probable imo.

The risk of Solution B can be realized when there is a *transient* server failure, where the NSS call passes, but the following PAM call fails. Given that both calls are retried, and the already restricting prerequisite of it, I think it will be very unlikely to happen in practice. And if this is only a regular user, the damage of the potential bad login won't be too dramatic either.

So I am leaning towards Solution B (current implementation). WDYT?

google_compute_engine_oslogin/authorized_keys/authorized_keys.cc
wonderfly: Yeah, I'd confirm that. If that's true, 200 with an empty body means something else then. Also, the 404 alone probably can't tell us whether it was because the user didn't exist, or the URL we used with curl was simply wrong. The latter is unlikely as it can be discovered and fixed in testing, so maybe we could assume 404 means "User Not Found". :)

illfelder:
The solution in my comment above doesn't prevent local user login, and prevents OS Login users from logging in without an authorization check. That obviates the security concern without affecting availability. The only downside is a file creation for every user.

PR #451, opened 2017-08-08
- Add separators

rjschwei:
+ Lets not assume that the user is diligent with adding separators
    for directory assignments, especially for paths that are composed
    internally in the Makefile. We may end up with double separators,
    but the OS gnores those.


wrigri:
So, I don't think adding the separators generally harms anything, and I'm not an expert in Makefile creation, but every single example of $DESTDIR and $PREFIX I've found in Makefiles has included these values without separators, including the GNU Coding standards: https://www.gnu.org/prep/standards/html_node/DESTDIR.html

So, I'm not necessarily opposed to this change, but it seems to go against the standard way these values tend to be used.

rjschwei:
The issue is really with the third component of the path, i.e. $(NSS_INSTALL_PATH), $(PAM_INSTALL_PATH) etc.

In rpm build these get determined via the %{_lib} macro to handle the difference between 32 and 64 bit, i.e. expand to lib or lib64, but does not expand to /lib and /lib64. Added the additional separators as "safeguards" but they are not strctly necessary, thus I'll pull those.

PR #452, opened 2017-08-08
Deny login if we fail to get a server response.

Sarsate:
Default to PAM_PERM_DENIED during login if we get server failures.


madmod:
Is this going to be merged soon?

Failing open seems like a big security issue. This flaw might allow an attack like [this one against Duo 2FA](https://www.appsecconsulting.com/blog/leveraging-duo-securitys-default-configuration-to-bypass-two-factor-authent), but would be much more serious because it is the only single factor.

I think there should be a config option to fail open or closed which should default to closed. (So an admin can decide if the tradeoff of gaining availability at the cost of security makes sense for them.)

Thanks!

illfelder:
We mitigate the security risk in #454 so there is no security benefit associated with this pull request. We left this PR open in case the approach taken in #454 causes usability problems at a later time. In order avoid confusion, I'll close this RP off.

PR #454, opened 2017-08-15
Add local files to track os-login users in the event of a transient failure.

Sarsate:
Adds a local file when an os-login user successfully logs in. If a server error occurs, we will check if this file exists to deny os-login users, and permit local users.

Additionally, clean up HttpGet() to set an http response code to better respond to server errors.

### DIFF
--- a/google_compute_engine_oslogin/authorized_keys/authorized_keys.cc
+++ b/google_compute_engine_oslogin/authorized_keys/authorized_keys.cc
@@ -37,13 +37,16 @@ int main(int argc, char* argv[]) {
   std::stringstream url;
   url << kMetadataServerUrl << "users?username=" << UrlEncode(argv[1]);
   string user_response;
-  if (!HttpGet(url.str(), &user_response)) {
+  long http_code = 0;
+  if (!HttpGet(url.str(), &user_response, &http_code)) {
     return 1;
   }
-  if (user_response.empty()) {
+  if (http_code == 404) {
     // Return 0 if the user is not an oslogin user. If we returned a failure
     // code, we would populate auth.log with useless error messages.
     return 0;
+  } else if (user_response.empty() || http_code == 500) {
+    return 1;
   }
   string email = ParseJsonToEmail(user_response);
   if (email.empty()) {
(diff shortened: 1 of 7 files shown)


illfelder:
Thanks Danny!

google_compute_engine_oslogin/authorized_keys/authorized_keys.cc
```diff
@@ -37,13 +37,16 @@ int main(int argc, char* argv[]) {
...
+  long http_code = 0;
+  if (!HttpGet(url.str(), &user_response, &http_code)) {
     return 1;
   }
-  if (user_response.empty()) {
+  if (http_code == 404) {
     // Return 0 if the user is not an oslogin user. If we returned a failure
     // code, we would populate auth.log with useless error messages.
     return 0;
+  } else if (user_response.empty() || http_code == 500) {
```
Sarsate: Good point. I don't think we even need to differentiate anything other than 404 and 200.

google_compute_engine_oslogin/authorized_keys/authorized_keys.cc
```diff
@@ -58,7 +61,8 @@ int main(int argc, char* argv[]) {
   url << kMetadataServerUrl << "authorize?email=" << UrlEncode(email)
       << "&policy=login";
   string auth_response;
-  if (!HttpGet(url.str(), &auth_response) || auth_response.empty()) {
+  if (!HttpGet(url.str(), &auth_response, &http_code) || http_code > 400 ||
```
Sarsate: Yeah, we should not print keys for anything other than 200.

google_compute_engine_oslogin/nss_module/nss_oslogin.cc
```diff
@@ -57,7 +57,9 @@ int _nss_oslogin_getpwuid_r(uid_t uid, struct passwd *result, char *buffer,
   std::stringstream url;
   url << kMetadataServerUrl << "users?uid=" << uid;
   string response;
-  if (!HttpGet(url.str(), &response) || response.empty()) {
+  long http_code = 0;
+  if (!HttpGet(url.str(), &response, &http_code) || http_code >= 400 ||
```
Sarsate: We should probably also return an error for anything other than 200 here.

Sarsate:
I made a new PR to address these changes, PTAL! 

https://github.com/GoogleCloudPlatform/compute-image-packages/pull/466

PR #466, opened 2017-08-22
Strictly check for HTTP code 200.

Sarsate:
In many cases, we really should only proceed if we get http code 200.

This changes several of the checks to only accept 200. The only exception is in pam_oslogin_login.cc, which cares specifically about server failures.

### DIFF
--- a/google_compute_engine_oslogin/authorized_keys/authorized_keys.cc
+++ b/google_compute_engine_oslogin/authorized_keys/authorized_keys.cc
@@ -38,14 +38,13 @@ int main(int argc, char* argv[]) {
   url << kMetadataServerUrl << "users?username=" << UrlEncode(argv[1]);
   string user_response;
   long http_code = 0;
-  if (!HttpGet(url.str(), &user_response, &http_code)) {
-    return 1;
-  }
-  if (http_code == 404) {
-    // Return 0 if the user is not an oslogin user. If we returned a failure
-    // code, we would populate auth.log with useless error messages.
-    return 0;
-  } else if (user_response.empty() || http_code == 500) {
+  if (!HttpGet(url.str(), &user_response, &http_code) ||
+      user_response.empty() || http_code != 200) {
+    if (http_code == 404) {
+      // Return 0 if the user is not an oslogin user. If we returned a failure
+      // code, we would populate auth.log with useless error messages.
+      return 0;
+    }
     return 1;
   }
   string email = ParseJsonToEmail(user_response);
@@ -61,7 +60,7 @@ int main(int argc, char* argv[]) {
   url << kMetadataServerUrl << "authorize?email=" << UrlEncode(email)
       << "&policy=login";
   string auth_response;
-  if (!HttpGet(url.str(), &auth_response, &http_code) || http_code > 400 ||
+  if (!HttpGet(url.str(), &auth_response, &http_code) || http_code != 200 ||
       auth_response.empty()) {
     return 1;
   }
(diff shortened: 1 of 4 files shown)


illfelder:
Thanks! Can you add a short comment to each of the non-200 error code lines?

google_compute_engine_oslogin/pam_module/pam_oslogin_login.cc
```diff
@@ -58,9 +58,9 @@ PAM_EXTERN int pam_sm_acct_mgmt(pam_handle_t *pamh, int flags, int argc,
   url << kMetadataServerUrl << "users?username=" << UrlEncode(str_user_name);
   string response;
   long http_code = 0;
-  if (!HttpGet(url.str(), &response, &http_code) || http_code == 500) {
-    // First check if there is a local file indicating that this is an oslogin
-    // user that has logged in before.
+  // Check for server failures. In this case, we can check for a local file to
+  // see if the user is an oslogin user.
+  if (!HttpGet(url.str(), &response, &http_code) || http_code >= 500) {
```
wonderfly: Hmm, when we get a client error (anything between 201 to 499), it's probably a **programming bug**. But regardless, we still failed to get the user info from the server, and can't reliably tell if the user in question is an os-login user or not. In that case, shouldn't we fall back the to local user file as we do in the case of server errors?

In the current implementation, when we get a 3xx error for example, we'll skip the both if checks and go directly to `ParseJsonToEmail`, which will result in a crash: getting 3xx doesn't necessarily mean `response` will be empty, right?

So I think a more reliable logic should be:

* If we get a 200 with a non-empty response, great, it's an os-login user, go ahead and check its login permission
* If we get a 404, great still, it's *not* an os-login user, return PAM_SUCCESS
* In all other cases, we can't reliably tell, fall back to checking the local user's file

WDYT?

google_compute_engine_oslogin/authorized_keys/authorized_keys.cc
```diff
@@ -45,7 +45,7 @@ int main(int argc, char* argv[]) {
     // Return 0 if the user is not an oslogin user. If we returned a failure
     // code, we would populate auth.log with useless error messages.
     return 0;
-  } else if (user_response.empty() || http_code == 500) {
+  } else if (user_response.empty() || http_code != 200) {
     return 1;
```
wonderfly: nit: Like what you did in other modules, this probably can be merged with the above `if (!HttpGet())` check

google_compute_engine_oslogin/pam_module/pam_oslogin_login.cc
```diff
@@ -58,9 +58,9 @@ PAM_EXTERN int pam_sm_acct_mgmt(pam_handle_t *pamh, int flags, int argc,
...
     if (file_exists) {
       return PAM_PERM_DENIED;
     }
```
wonderfly: The `else if` check below kind of indicates that the server's behavior on non-existent users is not deterministic: it could either return 404, or not 404 but with an empty response. Doesn't sound like a reliable design. Is that true?

google_compute_engine_oslogin/pam_module/pam_oslogin_login.cc
Sarsate: That sounds good to me! It leaves little room for edge cases slipping through.

google_compute_engine_oslogin/pam_module/pam_oslogin_login.cc
Sarsate: It is not true, the server now always returns 404 for non-existent users. The check for empty response is leftover from testing in development. We should explicitly be checking for 404 here.

google_compute_engine_oslogin/authorized_keys/authorized_keys.cc
Sarsate: Done.

Sarsate:
Thanks for the comments! PTAL

wonderfly:
Thanks for making the changes! LGTM

PR #467, opened 2017-08-23
Update google-compute-engine from pypi in the SCL virtenv for EL6.

zmarano:


specs/google-compute-engine.spec
```diff
@@ -110,6 +110,11 @@ if [ $1 -eq 2 ]; then
   restart -q -n google-ip-forwarding-daemon
   /usr/bin/google_instance_setup
 fi
+
+# Install google-compute-engine from pypi into the SCL environment if it exists.
```
illfelder: Mind adding a little more detail to this comment? In particular that the path is specific to SCL and why we need to keep this package in sync with the boto installed on the VM?

zmarano: Done

PR #471, opened 2017-08-25
Add back hostname script to DHCP exit hooks.

zmarano:


debian/changelog
```diff
@@ -1,3 +1,9 @@
+google-compute-image-packages (2.6.0-2) stable; urgency=low
```
illfelder: Move the version bump to a separate pull request.

palladius:
Can you please explain why you did this change? It broke some customers workloads by rewriting `/etc/hosts`. I think this is a dangerous CL. Any chance we can revert it or make it togglable?

PR #474, opened 2017-08-28
Fix various package issues.

zmarano:


disk_expand/gce-disk-expand-el7.spec
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 Name: gce-disk-expand
 Summary: Google Compute Engine root disk expansion utilities for EL7
-Version: 1.0.3
+Version: 1.0.4
```
illfelder: Version bump in a separate pull request.

PR #476, opened 2017-09-08
Use curl to download metadata script files.

illfelder:
In Python 2.6, urllib does not check ssl certificates.


zmarano:
Add a requirement for curl to the package files:
https://github.com/GoogleCloudPlatform/compute-image-packages/blob/master/specs/google-compute-engine.spec#L34
https://github.com/GoogleCloudPlatform/compute-image-packages/blob/master/debian/control#L22

illfelder:
Added dependency.

wonderfly:
lgtm

PR #477, opened 2017-09-14
Match any barrier option.

zmarano:


disk_expand/gce-disk-expand-el7.spec
```diff
@@ -44,7 +44,7 @@ cp -R $RPM_SOURCE_DIR/usr $RPM_BUILD_ROOT
 %post
 %systemd_post expand-root.service
 # Remove barrier options in fstab for EL7.
-sed -i 's/defaults,barrier=1/defaults/' /etc/fstab
+sed -i 's/defaults,barrier=./defaults/' /etc/fstab
```
illfelder: `\d` will match any number which might be better. WDYT?

zmarano: Sure, that works.

illfelder: Since that apparently doesn't work in the bash version of sed, you can use this: `defaults,barrier[^ ]*` to remove all characters after barrier until the next space.

PR #479, opened 2017-09-14
Bump the Python package version to 2.6.1.

illfelder:


specs/google-compute-engine.spec
```diff
@@ -18,7 +18,7 @@
 %endif
 
 Name: google-compute-engine
-Version: 2.6.0
+Version: 2.6.1
 Release: 3%{?dist}
```
zmarano: 1%{?dist}

illfelder: Done.

PR #480, opened 2017-09-15
Fix system hang when shutown

pigfoot:
Existing systemd shtudown script would cause system hang by waiting for DNS response but system-resolved service may be down.


pigfoot:
I signed it!

pigfoot:
@zmarano got it, will do. Thanks for the reminder.

PR #484, opened 2017-09-28
Remove broken/redundant rm command from control script.

Sarsate:
Removes broken/redundant rm command from control script.


Sarsate:
I signed it!

Sarsate:
I signed it!

PR #486, opened 2017-09-29
Added udev rules to symlink nvme devices into the disk/by-id/ directory

### LINKED ISSUE #485: Create udev rule for nvme disks to /dev/disk/by-id symlinks
A rule needs to be created so that uniquely identifiable symlinks are created for all attached nvme storage devices. This is necessary for us to enable nvme local ssd storage for users.

Maybe added to this file:
https://github.com/GoogleCloudPlatform/compute-image-packages/blob/development/google_config/udev/65-gce-disk-naming.rules

davidz627:
Fixes #485 

We need to be able to access nvme devices by unique id in the disk/by-id directory. I'm not that familiar with udev rules so please take a close look at this.

cc @msau42

### DIFF
--- a/google_config/udev/65-gce-disk-naming.rules
+++ b/google_config/udev/65-gce-disk-naming.rules
@@ -20,5 +20,7 @@ SUBSYSTEM!="block", GOTO="gce_disk_naming_end"
 KERNEL=="sd*|vd*", IMPORT{program}="scsi_id --export --whitelisted -d $tempnode"
 KERNEL=="sd*|vd*", ENV{ID_SERIAL_SHORT}=="?*", ENV{DEVTYPE}=="disk", SYMLINK+="disk/by-id/google-$env{ID_SERIAL_SHORT}"
 KERNEL=="sd*|vd*", ENV{ID_SERIAL_SHORT}=="?*", ENV{DEVTYPE}=="partition", SYMLINK+="disk/by-id/google-$env{ID_SERIAL_SHORT}-part%n"
+KERNEL=="nvme*", ENV{ID_SERIAL_SHORT}=="?*", ENV{DEVTYPE}=="disk", SYMLINK+="disk/by-id/google-%k"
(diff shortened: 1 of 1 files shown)


davidz627:
cc @zmarano @illfelder

davidz627:
The contents of /dev/disk/by-id/ with 8 NVMe SSD's attached with nvme0n1 having an extra primary partition.

lrwxrwxrwx 1 root root 13 Sep 29 17:58 google-nvme0n1 -> ../../nvme0n1
lrwxrwxrwx 1 root root 15 Sep 29 17:58 google-nvme0n1p1 -> ../../nvme0n1p1
lrwxrwxrwx 1 root root 13 Sep 29 17:58 google-nvme0n2 -> ../../nvme0n2
lrwxrwxrwx 1 root root 13 Sep 29 17:58 google-nvme0n3 -> ../../nvme0n3
lrwxrwxrwx 1 root root 13 Sep 29 17:58 google-nvme0n4 -> ../../nvme0n4
lrwxrwxrwx 1 root root 13 Sep 29 17:58 google-nvme0n5 -> ../../nvme0n5
lrwxrwxrwx 1 root root 13 Sep 29 17:58 google-nvme0n6 -> ../../nvme0n6
lrwxrwxrwx 1 root root 13 Sep 29 17:58 google-nvme0n7 -> ../../nvme0n7
lrwxrwxrwx 1 root root 13 Sep 29 17:58 google-nvme0n8 -> ../../nvme0n8

illfelder:
Thanks for the contribution! Please make the pull request against the development branch instead of master.

davidz627:
updated

davidz627:
ok i've kind of messed up the commit. not really sure what happened. attempting fix now

zmarano:
Can you verify this works on CentOS/RHEL as well as Debian/Ubuntu?

msau42:
@zmarano one question, this udev rule creates the symlink as google-nvme0n1, which is inconsistent from the naming for scsi local SSDs.   When we tried using the same udev rule as scsi for nvme, we found that the $env{ID_SERIAL_SHORT} is just "nvme_card", which gets overridden by subsequent nvme local SSDs.  Is there perhaps another underlying issue that needs to be fixed for nvme local SSDs?

PR #487, opened 2017-10-05
Allow string type for uids, gids, and expiration times.

Sarsate:


google_compute_engine_oslogin/utils/oslogin_utils.cc
```diff
@@ -338,17 +338,19 @@ bool ParseJsonToPasswd(string response, struct passwd* result,
     string string_key(key);
 
     if (string_key == "uid") {
-      if (val_type != json_type_int) {
+      if (val_type == json_type_int || val_type == json_type_string) {
```
illfelder: Based on these changes, we have a bug in the code above (line 278) where we don't currently respect expiration times; all `int64` values are written as strings in json.

Also, please run the sed expression `s/OSLogin/OS Login/g`.

Sarsate: Done. I still need to do some testing and make this resilient to strings that aren't parsable to and integer

Sarsate: The parse will return 0 if the value malformed. I added code that will return EINVAL if we detect this for uid and gid. For expiration keys, it is inconsequential since an expiration time of 0 will always be expired.

I also added some unit tests for the gid/uid cases.

illfelder: When gid is set to 0, let's set the gid to the uid value. That will allow users where a gid is not set, and prevent accidentally adding a user with root access.

PR #503, opened 2017-10-18
Fix Debian packaging to correctly restart services.

zmarano:


illfelder:
Awesome job getting package updates working properly. It looks like package removal still doesn't work (it doesn't stop running daemons), and I think package install doesn't start daemons, but I'm not sure if that is a problem.

Issue #504, opened 2017-10-18
instance_setup should generate SSH host keys when none are present


jkaplowitz:
Right now, instance_setup does not correctly handle the case where no SSH host keys are present, failing to generate them and preventing SSH from starting.

Instead, when instance_setup triggers a key (re)generation, it could work this way:

1. If instance_config specifies a list of cipher suites for which to generate host keys, generate that list.
1. Otherwise, if any host keys are present on the system, generate that list.
1. Otherwise, unconditionally generate ed25519, ecdsa, and rsa host keys. Don't generate dsa keys in this case, as they're no longer viewed as secure; for example, Debian 9 no longer generates these keys upon a fresh install.

In terms of evaluating success vs failure, a success would be "at least one freshly generated host key is created, and no pre-existing host key remains on the system." If no host keys end up on the disk, it willl be impossible to log in, and if any old host keys remain present, there may be a security issue.

I suggest logging a non-blocking debugging message if any attempted cipher types fail to generate, but this is not inherently a problem by itself.

illfelder:
A fix is merged into development. The change should make it into our next release if all goes well.

illfelder:
We released this change a while back. Closing off the issue.

PR #510, opened 2017-10-27
Generate SSH host keys when none are present.

illfelder:
Provide an instance config option for specifying which host key types to
generate.

### DIFF
--- a/google_compute_engine/instance_setup/tests/instance_setup_test.py
+++ b/google_compute_engine/instance_setup/tests/instance_setup_test.py
@@ -49,6 +49,7 @@ class InstanceSetupTest(unittest.TestCase):
     mock_watcher.MetadataWatcher.return_value = mock_watcher_instance
     mock_config_instance = mock.Mock()
     mock_config_instance.GetOptionBool.return_value = True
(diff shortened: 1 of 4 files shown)


illfelder:
Fixes #504.

google_compute_engine/instance_setup/tests/instance_setup_test.py
```diff
@@ -325,21 +328,23 @@ def testSetSshHostKeysFirstBoot(self, mock_listdir):
     self.mock_setup._GenerateSshKey = mock_generate_key
     mock_listdir.return_value = [
         'ssh_config',
-        'ssh_host_rsa_key',
+        'ssh_host_dsa_key',
```
zmarano: Should be ssh_host_ecdsa_key if we are doing that by default.
Older distros still have dsa keys (EL6, Deb8)

illfelder: These aren't default, this is a made up list of strings for testing. The types may as well be "apple" "puppy" "cookie". This is just verifying that the existing files on the system (with names different from the names in the config list) will also be regenerated.

PR #524, opened 2017-11-28
Update build to use json-c's recommended includedir.

Sarsate:
Changes inclusion of json-c's headers to #include <json.h> and updates
the build to add /usr/include/json-c to the search path.


wonderfly:
LGTM. Thanks for the quick change.

PR #537, opened 2018-01-25
Bump package versions.

zmarano:


google_compute_engine_oslogin/packaging/debian8/changelog
```diff
@@ -1,3 +1,9 @@
+google-compute-engine-oslogin (1.1.3-1+deb8) unstable; urgency=low
+
+  * Change the OS Login uid restriction to < 1000.
```
illfelder: I would update this to: `Change the OS Login uid restriction to allow uid 1000`

PR #538, opened 2018-01-26
oslogin: Add pam config differently on COS

wonderfly:
Such that it doesn't depend on the existence of a certain string in the
sshd pam config.


wrigri:
So a couple of concerns:
1. "local -r" doesn't work with /bin/sh on Debian (may also be a problem on other distros, but I didn't check all of them)
2. On Debian (and other distros that support SELinux), there is the following comment in /etc/pam.d/sshd:
# SELinux needs to be the first session rule.  This ensures that any
# lingering context has been cleared.  Without this it is possible that a
# module could execute code in the wrong domain.

This change adds the oslogin session rule before that, which doesn't seem to be where we would want it.

wrigri:
Sorry about the bold. It was a cut and paste comment line and the "#" got interpreted by the markdown.

wonderfly:
No worries. The selinux thing is a good point. I've pushed another commit that does this only for COS. PTAL.

illfelder:
I've updated the base to diff against development, but please update your branch to remove any potential accidental changes.

wonderfly:
I created https://github.com/GoogleCloudPlatform/compute-image-packages/pull/542 instead, to diff against the development branch.

PR #539, opened 2018-01-26
Fix hostname setting and always set the transient hostname.

zmarano:


kladiv:
+1
`google_set_hostname` still miss _hostnamectl_ code block. Tested on CentOS7 (the same in CentOS8, etc.) with no NetworkManager.

```
hostnamectl=$( which hostnamectl 2> /dev/null )
  if [ -x "$hostnamectl" ]; then
    $hostnamectl set-hostname "$new_host_name"
```

Static hostname (/etc/hostname) is still not changed and appears equal to VM Image.

I see _hostnamectl_ code block is still in development tree.
Need in production

Best Regards,
Claudio

PR #540, opened 2018-01-27
Close socket connections when requesting metadata.

illfelder:
Currently we are not cleaning up properly after OS Login related
metadata queries resulting in open socket connections.


google_compute_engine_oslogin/utils/oslogin_utils.cc
```diff
@@ -171,14 +172,17 @@ bool HttpGet(const string& url, string* response, long* http_code) {
 
       code = curl_easy_perform(curl);
       if (code != CURLE_OK) {
+        curl_global_cleanup();
+        curl_easy_cleanup(curl);
```
Sarsate: here and below, curl_easy_cleanup() should be before global_cleanup()

Sarsate: and above :)

illfelder: Yep, thanks!

Sarsate:
LGTM!

PR #542, opened 2018-01-29
oslogin: Add pam config differently on COS

wonderfly:
The way pam is configured for sshd on COS is different from other
distros; it's got nothing but four `include`s that point to another
config. So the insertion of oslogin config has to be done differently.


illfelder:
LGTM

PR #547, opened 2018-01-31
Muti-Nic network setup support for SUSE 11 and 12.

ryanwe:
Adds network setup logic to handle multiple Nics for SUSE 11 and 12.

For SUSE 11
- Run `dhcpcd` on additional Nics.

For SUSE 12
- For additional Nics, create the `ifcfg-eth*` files.
- Run `wicked ifup eth1...` to active the Nics.

### DIFF
--- a/google_compute_engine/compat.py
+++ b/google_compute_engine/compat.py
@@ -40,6 +40,8 @@ elif 'debian' in distro_name and distro_version == '9':
   import google_compute_engine.distro.debian_9.utils as distro_utils
 elif 'suse' in distro_name and distro_version == '12':
   import google_compute_engine.distro.sles_12.utils as distro_utils
+elif 'suse' in distro_name and distro_version == '11':
+  import google_compute_engine.distro.sles_11.utils as distro_utils
 else:
   # Default to Debian 9.
   import google_compute_engine.distro.debian_9.utils as distro_utils
--- a/google_compute_engine/compat.py
+++ b/google_compute_engine/compat.py
@@ -16,31 +16,35 @@
 """A module for resolving compatibility issues between Python 2 and Python 3."""
 
 import logging
-import platform
 import subprocess
 import sys
 
-# Set distro-specific utils.
-distribution = platform.linux_distribution()
+# Get the linux distribution.
+if sys.version_info >= (3.5, 0):
+  import distro as distro  # Must be installed on the image.
+else:
+  import platform as distro
+
+distribution = distro.linux_distribution()
 distro_name = distribution[0].lower()
 distro_version = distribution[1].split('.')[0]
 distro_utils = None
 
 if 'centos' in distro_name and distro_version == '6':
   import google_compute_engine.distro.el_6.utils as distro_utils
-elif 'centos' in distro_name and distro_version == '7':
+elif 'centos' in distro_name:
   import google_compute_engine.distro.el_7.utils as distro_utils
 elif 'red hat enterprise linux' in distro_name and distro_version == '6':
   import google_compute_engine.distro.el_6.utils as distro_utils
-elif 'red hat enterprise linux' in distro_name and distro_version == '7':
+elif 'red hat enterprise linux' in distro_name:
   import google_compute_engine.distro.el_7.utils as distro_utils
 elif 'debian' in distro_name and distro_version == '8':
   import google_compute_engine.distro.debian_8.utils as distro_utils
-elif 'debian' in distro_name and distro_version == '9':
+elif 'debian' in distro_name:
   import google_compute_engine.distro.debian_9.utils as distro_utils
 elif 'suse' in distro_name and distro_version == '11':
   import google_compute_engine.distro.sles_11.utils as distro_utils
-elif 'suse' in distro_name and distro_version == '12':
+elif 'suse' in distro_name:
   import google_compute_engine.distro.sles_12.utils as distro_utils
 else:
   # Default to Debian 9.
--- a/google_compute_engine/compat.py
+++ b/google_compute_engine/compat.py
@@ -19,9 +19,8 @@ import logging
 import subprocess
 import sys
 
-# Get the linux distribution.
-if sys.version_info >= (3.5, 0):
-  import distro as distro  # Must be installed on the image.
+if sys.version_info >= (3, 5):
+  import distro
 else:
   import platform as distro
 
--- a/google_compute_engine/compat.py
+++ b/google_compute_engine/compat.py
@@ -38,6 +38,8 @@ elif 'debian' in distro_name and distro_version == '8':
   import google_compute_engine.distro.debian_8.utils as distro_utils
 elif 'debian' in distro_name and distro_version == '9':
   import google_compute_engine.distro.debian_9.utils as distro_utils
+elif 'suse' in distro_name and distro_version == '12':
+  import google_compute_engine.distro.sles_12.utils as distro_utils
 else:
   # Default to Debian 9.
   import google_compute_engine.distro.debian_9.utils as distro_utils
--- a/google_compute_engine/compat.py
+++ b/google_compute_engine/compat.py
@@ -38,10 +38,10 @@ elif 'debian' in distro_name and distro_version == '8':
   import google_compute_engine.distro.debian_8.utils as distro_utils
 elif 'debian' in distro_name and distro_version == '9':
   import google_compute_engine.distro.debian_9.utils as distro_utils
-elif 'suse' in distro_name and distro_version == '12':
-  import google_compute_engine.distro.sles_12.utils as distro_utils
 elif 'suse' in distro_name and distro_version == '11':
   import google_compute_engine.distro.sles_11.utils as distro_utils
+elif 'suse' in distro_name and distro_version == '12':
+  import google_compute_engine.distro.sles_12.utils as distro_utils
 else:
   # Default to Debian 9.
   import google_compute_engine.distro.debian_9.utils as distro_utils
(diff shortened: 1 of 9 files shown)


google_compute_engine/distro/sles_11/utils.py
```diff
@@ -0,0 +1,60 @@
...
+      interfaces: list of string, the output device names to enable.
+      logger: logger object, used to write to SysLog and serial port.
+    """
+    for interface in interfaces:
+      dhcpcd = ['/sbin/dhcpcd']
+      try:
+        subprocess.check_call(dhcpcd + ['-x'] + [interface])
+      except subprocess.CalledProcessError:
+        # Dhcpcd not yet running for this device.
+        logger.info('Could not exit interface %s.', interface)
```
illfelder: I'm not sure what this message means. If its problematic it should be a warning. If it's expected it should be less concerning (and more clear).

google_compute_engine/compat.py
```diff
@@ -38,6 +38,10 @@
   import google_compute_engine.distro.debian_8.utils as distro_utils
 elif 'debian' in distro_name and distro_version == '9':
   import google_compute_engine.distro.debian_9.utils as distro_utils
+elif 'suse' in distro_name and distro_version == '11':
+  import google_compute_engine.distro.sles_11.utils as distro_utils
+elif 'suse' in distro_name and distro_version == '12':
```
illfelder: It might be better to remove the `and distro_version == '12'`. That way other suse versions will run code that has some chance of working.

illfelder:
/cc @rjschwei 

Hi Robert, please take a look and verify we are using the recommended approach for enabling multiple network interfaces on SLES images.

google_compute_engine/compat.py
rjschwei: agreed, without version test for 12, openSUSE Leap 42.3 will work as well as the upcoming versions openSUSE Leap 15 and SLE 15.

google_compute_engine/compat.py
```diff
@@ -38,6 +38,10 @@
...
+  import google_compute_engine.distro.sles_12.utils as distro_utils
```
rjschwei: Probably rename this to suse_wicked

google_compute_engine/distro/sles_12/utils.py
```diff
@@ -0,0 +1,84 @@
...
+      interface_config = os.path.join(
+          self.network_path, 'ifcfg-%s' % interface)
+      interface_content = [
+          '# Added by Google.',
+          'STARTMODE=hotplug',
+          'BOOTPROTO=dhcp',
+          'DHCLIENT_SET_DEFAULT_ROUTE=yes',
+          'DHCLIENT_ROUTE_PRIORITY=10%s00' % interface,
+          'CLOUD_NETCONFIG_MANAGE=yes',
+          'POST_DOWN_SCRIPT=compat:suse:cloud-netconfig-cleanup',
```
rjschwei: @jgleissner given the interfaces are setup in the initialization code would we still run the cleanup code from cloud-net-config? Also we'd need to build a generic cloud-netconfig package that provides this script.

google_compute_engine/distro/sles_11/utils.py
ryanwe: The `-x` option is to "exit" the interface which is basically turning it off. I went with dhcpcd's wording, but I'll change it so it's clearer.

google_compute_engine/compat.py
illfelder: I'd actually prefer we leave it as is. We might want to extend the functionality of the distro configs to extend beyond networking.

rjschwei: If it extends much beyond networking it might be worth to push version decisions into the distro modules and thus just load a suse, debian, rhel, etc module and then call central entry points for interface_config and whatever else might be needed. Then the complexity of the version decision is pushed to a lower level and it would eliminate what will be an ever increasing if-elif-else block in this file.

illfelder: Ack - we can refactor this as it becomes unwieldy. Currently it seems manageable.

google_compute_engine/distro/sles_12/utils.py
jgleissner: @rjschwei cloud-netconfig-cleanup removes routing policies associated with the interface that was brought down. cloud-netconfig creates routing policies to make sure packets are sent out through the interface associated with the source IP address. AFACIS the initialization code in this project doesn't do that, so this script would not be needed, unless cloud-netconfig is intended to be used. Which would make sense actually, as it creates the aforementioned routing policies and also can configure secondary (alias) IP addresses, unless that functionality is intended to be replicated in this initialization code.

rjschwei: @jgleissner yes, I guess the question if for @ryanwe and @illfelder to answer, what is the expectation how the routing policies for multiple network interfaces are handeled? Does the GCE DHCP server provide the routing information for the new interface, i.e. the policy is expected to be added by the distribuiton dhcp client? I the answer is yes, then

```
'CLOUD_NETCONFIG_MANAGE=yes',
'POST_DOWN_SCRIPT=compat:suse:cloud-netconfig-cleanup',
```
should be removed. The current implementation of cloud-netconfig and the packages we build extract routing information from the metadata in EC2 and Azure and set up routing, as @jgleissner pointed out.

The code would indicate that the answer to my question is "yes, dhcp client gets the route information from the GCE DHCP server and is expected to set he routes", but I want to double check tha this is correct. Thanks

ryanwe: @rjschwei, yes that is correct. I'll remove those lines. I also updated our distribution checks to use `distro`. Thanks!

google_compute_engine/compat.py
```diff
@@ -16,28 +16,36 @@
...
 
 import logging
-import platform
 import subprocess
 import sys
 
-# Set distro-specific utils.
-distribution = platform.linux_distribution()
+# Get the linux distribution.
+if sys.version_info >= (3.5, 0):
```
illfelder: Your version checking should be `(3, 5)` instead of `(3.5, 0)`.

google_compute_engine/compat.py
```diff
@@ -16,28 +16,36 @@
...
+  import distro as distro  # Must be installed on the image.
```
illfelder: Leave out the `as distro` portion (and the comment isn't needed).

google_compute_engine/compat.py
```diff
@@ -16,28 +16,36 @@
 """A module for resolving compatibility issues between Python 2 and Python 3."""
 
 import logging
-import platform
 import subprocess
 import sys
 
-# Set distro-specific utils.
-distribution = platform.linux_distribution()
+# Get the linux distribution.
```
illfelder: Please either remove this comment or move it (and update the wording) to the section where you are importing configs specific to the Linux distribution.

google_compute_engine/distro/sles_11/utils.py
```diff
@@ -0,0 +1,60 @@
...
+    """Use dhcpcd to activate the interfaces.
+
+    Args:
+      interfaces: list of string, the output device names to enable.
+      logger: logger object, used to write to SysLog and serial port.
+    """
+    for interface in interfaces:
+      dhcpcd = ['/sbin/dhcpcd']
+      try:
+        subprocess.check_call(dhcpcd + ['-x'] + [interface])
```
illfelder: `subprocess.check_call(dhcpcd + ['-x', interface])`

google_compute_engine/distro/sles_12/tests/utils_test.py
```diff
@@ -0,0 +1,94 @@
...
+
+  @mock.patch(
+      'google_compute_engine.distro.sles_12.utils.subprocess.check_call')
+  def testIfup(self, mock_call):
+    mocks = mock.Mock()
+    mocks.attach_mock(mock_call, 'call')
+    mocks.attach_mock(self.mock_logger, 'logger')
+    mock_call.side_effect = [
+        None,
+        subprocess.CalledProcessError(1, 'Test'),
```
illfelder: Move this up a line next to the `None`.

google_compute_engine/distro/sles_12/utils.py
jgleissner: The routing policies set up by cloud-netconfig are actually not based on metadata information, at least not directly. cloud-netconfig creates a routing policy for any IP address (i.e. also the one set by DHCP) it finds on any secondary interfaces. This is useful, as the kernel will pick the default route via eth0 when sending packets not directly destined for the network associated with i.e. eth1, even when their source IP address is on that network. So for instance, if you start an instance with two network interfaces, with both having an external IP associated, you can see that you can ssh into the machine on the one associated with eth0 but not on the eth1 one. Using policy routing amends this, which is why we created cloud-netconfig to automatically do that. So the question is whether the GCE initialization code is going to do this.

The metadata information are used by cloud-netconfig, as mentioned, to configured secondary IP addresses in EC2 and Azure. I thought "alias IP" would be GCE's equivalent, but I misunderstood that. It's actually an additional IP range the GCE will route to the VM, probably intended for use with container hosts. So for cloud-netconfig, should we want to use it on GCE, there's nothing really to do at this point. It could set up routing policies for those ranges, but since alias IP ranges are only supported with single-nic instances, it's not necessary.

google_compute_engine/compat.py
ryanwe: removed :hocho:

illfelder:
You'll need to update `.travis.yml` to include distro.

illfelder:
You should also probably update `setup.py`.

Something like this:
```
install_requires=['boto', 'setuptools', 'distro;python_version>="3.5"'],
```

PR #560, opened 2018-02-22
Fix setup.py dependencies.

zmarano:


setup.py
```diff
@@ -17,16 +17,21 @@
 
 import glob
 
+import sys
 import setuptools
 
 
+install_requires = ['boto', 'setuptools']
+if sys.version_info >= (3,5):
```
illfelder: Here, and in all parts of the English language, space after comma.

setup.py
```diff
@@ -17,16 +17,21 @@
...
+  install_requires = install_requires + ['distro']
```
illfelder: This should be +=.

PR #561, opened 2018-02-22
Prototype for nss_cache binary. DO NOT MERGE

Sarsate:
This is a prototype PR for a simple nss_cache. DO NOT MERGE THIS PR. I'm only sending this out to make it easier to locate and mess with.


google_compute_engine_oslogin/utils/oslogin_utils.h
```diff
@@ -79,6 +79,12 @@ class NssCache {
   // true on success.
   bool LoadJsonArrayToCache(string response);
 
+  // Helper method that effectively implements the getpwent_r nss method. Each call will
```
illfelder: Line length should be at most 80 characters long (unless I'm missing something).

wrigri:
When trying to build the latest version of this, I got:
nss_cache/nss_cache.cc: In function ‘int main(int, char**)’:
nss_cache/nss_cache.cc:61:42: error: ‘buflen’ was not declared in this scope
     BufferManager buffer_manager(buffer, buflen);
                                          ^~~~~~

It looks like the 'int buflen = 2048' line was removed.

Sarsate:
Thanks for catching that, should be fixed now!

wrigri:
LGTM

Sarsate:
I am making some minor changes to this. Don't merge yet.

Sarsate:
Ok PTAL. The binary will create a backup in case caching fails, and attempt to restore the cache on failure. There are also some changes to the utils file, since I found a discrepancy between what the library expected for page_token behavior, and what actually happens in production.

The library thought that the last page would contain an empty page token, along with the final array of passwd entries. What actually happens is the last page of actual passwd entries still has a page token, and the *next* page will contain no passwd entries, but a page token of 0.

wrigri:
LGTM
Seems to work as expected.

PR #563, opened 2018-02-26
Clear the CURL_GLOBAL_SSL bit on curl initialization. Fixes #562.

### LINKED ISSUE #562: oslogin breaks OpenSSL setup
Hi,

The problem has been shown that when we use `getpwent()` with oslogin nss module, SSL setup will be broken after the call.

Here is a simple code to reproduce it.
```c
#include <stdio.h>
#include <openssl/evp.h>
#include <pwd.h>

int main() {
    OpenSSL_add_all_algorithms();
    EVP_get_digestbyname("sha256") ? printf("loaded\n") : printf("not loaded\n");

    struct passwd *pw;
    setpwent();
    while ((pw = getpwent()) != 0) {}

    EVP_get_digestbyname("sha256") ? printf("loaded\n") : printf("not loaded\n");
    return 0;
}
```
```
$ gcc test-getpwent.c -lssl -lcrypto -o test-getpwent
$ ./test-getpwent
loaded
not loaded

# after removing oslogin from /etc/nsswitch.conf
$ ./test-getpwent
loaded
loaded
```

After tracking it down, I found that the cause can be the `curl_global_cleanup()` called by `oslogin_utils::HttpGet`.

As `oslogin` communicates with metadata without SSL/TLS, can we clear the `CURL_GLOBAL_SSL` bit when initialize curl so that it will not cleanup OpenSSL things?

A simple test is as follows.
```c
#include <stdio.h>
#include <openssl/evp.h>
#include <curl/curl.h>

int main() {
    OpenSSL_add_all_algorithms();
    EVP_get_digestbyname("sha256") ? printf("loaded\n") : printf("not loaded\n");

    curl_global_init(CURL_GLOBAL_ALL & ~CURL_GLOBAL_SSL);
    curl_global_cleanup();

    EVP_get_digestbyname("sha256") ? printf("loaded\n") : printf("not loaded\n");
    return 0;
}
```
```
$ gcc test-curl.c -lssl -lcrypto -lcurl -o test-curl
$ ./test-curl
loaded
loaded
```

Test environment:
```
libcurl3/xenial-updates,xenial-security,now 7.47.0-1ubuntu2.6 amd64 [installed]
libcurl3-gnutls/xenial-updates,xenial-security,now 7.47.0-1ubuntu2.6 amd64 [installed]
libcurl4-openssl-dev/xenial-updates,xenial-security,now 7.47.0-1ubuntu2.6 amd64 [installed]
openssl/xenial-updates,xenial-security,now 1.0.2g-1ubuntu4.10 amd64 [installed]
```

What do you think?

By the way, the original problem was shown when setting up instances via Chef on GCE with Ubuntu 16.04.  Ruby initializes the OpenSSL when requiring `openssl` (`Init_openssl`) and Chef calls `Etc.passwd` somewhere which leads to a runtime error when trying to use openssl again. 
```
FATAL: RuntimeError: cookbook_file[xxxx] (xxx::xxx line N) had an error: RuntimeError: Unsupported digest algorithm (SHA256).
```

Sarsate:
This was breaking OpenSSL setup. Since oslogin does not use SSL while
communicating with the metadata server, we can clear this bit on
initialization. Fixes #562.

### DIFF
--- a/google_compute_engine_oslogin/utils/oslogin_utils.cc
+++ b/google_compute_engine_oslogin/utils/oslogin_utils.cc
@@ -149,7 +149,7 @@ bool HttpGet(const string& url, string* response, long* http_code) {
     return false;
   }
   CURLcode code(CURLE_FAILED_INIT);
-  curl_global_init(CURL_GLOBAL_ALL);
+  curl_global_init(CURL_GLOBAL_ALL & ~CURL_GLOBAL_SSL);
   CURL* curl = curl_easy_init();
   std::ostringstream response_stream;
   int retry_count = 0;


wonderfly:
Can you add `fixed #562` to the PR description?

PR #564, opened 2018-02-26
Bump OS Login package version.

zmarano:


wonderfly:
Please hold off this release as it might not solve the problem on newer curl versions.
See https://github.com/GoogleCloudPlatform/compute-image-packages/issues/562#issuecomment-368627208

PR #565, opened 2018-02-26
Add module blacklist for GCE.

zmarano:


google_config/modprobe/gce-blacklist.conf
```diff
@@ -0,0 +1,5 @@
+# nouveau does not work with GCE GPU's
```
illfelder: Please add a period to the end of your comments.

PR #570, opened 2018-03-05
Accounts daemon updates the OS Login NSS cache.

illfelder:
This includes testing for updates to the OS Login NSS cache.


illfelder:
Blocked by #561.

PR #582, opened 2018-03-30
Update udev rules to name and symlink NVMe disks and partitions

### LINKED ISSUE #485: Create udev rule for nvme disks to /dev/disk/by-id symlinks

anilyigitfiliz:
Fixes #485

Description:
We don't have custom naming/symlink rules for NVMe disks which results in them not getting unique symlinks under /dev/disk/by-id (when you have multiple NVMe disks all you get is a single nvme_card entry).

Fix:
Update the custom GCE udev rules file and add rules for NVMe disks. With this change, the NVMe disks will have symlinks (and ID_SERIAL and ID_SERIAL_SHORT attributes in case anyone relies on them) similar to those of SCSI disks.

Test Notes:
Verified that this works for CentOS 6, Debian 7, 8, 9, Ubuntu 16 and SUSE 12.

### DIFF
--- a/google_config/udev/65-gce-disk-naming.rules
+++ b/google_config/udev/65-gce-disk-naming.rules
@@ -21,14 +21,14 @@ SUBSYSTEM!="block", GOTO="gce_disk_naming_end"
 KERNEL=="sd*|vd*", IMPORT{program}="scsi_id --export --whitelisted -d $tempnode"
 
 # NVME naming
-KERNEL=="nvme0n1", ENV{ID_SERIAL_SHORT}="local-ssd-0"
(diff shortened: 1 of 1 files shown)


illfelder:
Please verify other team member's internally are happy with this change and let me know when you're confident we can merge. From my end this looks good to me.

PR #586, opened 2018-04-25
Import the minimum set of libnss-cache and change the cache file name

gzhwei:
Import the minimum set of libnss-cache and change the cache file name to be oslogin specific


illfelder:
This looks great! I think we need to update the license headers, but other than that I don't see any issues.

gzhwei:
Thanks @Sarsate for the formatting suggestions!

PR #592, opened 2018-05-03
Add HandleClockSync in distro/utils.py for FreeBSD

helen-fornazier:
There is no hwclock for FreeBSD, use ntpdate instead.

- Check in compat.py if the system is FreeBSD and don't call
distro.linux_distribution() if true
- Migrate distro specific logic for clock sync to distro/utils.
- Add the following functions to distro/helpers:
        def CallHwclock(logger): # Logic used by Linux systems
        def CallNtpdate(logger): # Logic used by FreeBSD
- Add mock tests accordingly

### DIFF
--- a/google_compute_engine/clock_skew/tests/clock_skew_daemon_test.py
+++ b/google_compute_engine/clock_skew/tests/clock_skew_daemon_test.py
@@ -71,36 +71,22 @@ class ClockSkewDaemonTest(unittest.TestCase):
       ]
       self.assertEqual(mocks.mock_calls, expected_calls)
 
-  @mock.patch('google_compute_engine.clock_skew.clock_skew_daemon.subprocess.check_call')
-  def testHandleClockSync(self, mock_call):
-    command = ['/sbin/hwclock', '--hctosys']
+  @mock.patch('google_compute_engine.clock_skew.clock_skew_daemon.distro_utils')
+  def testHandleClockSync(self, mock_distro_utils):
     mock_sync = mock.create_autospec(clock_skew_daemon.ClockSkewDaemon)
     mock_logger = mock.Mock()
     mock_sync.logger = mock_logger
+    mock_sync.distro_utils = mock_distro_utils
 
     clock_skew_daemon.ClockSkewDaemon.HandleClockSync(mock_sync, 'Response')
-    mock_call.assert_called_once_with(command)
     expected_calls = [
         mock.call.info(mock.ANY, 'Response'),
-        mock.call.info(mock.ANY),
     ]
     self.assertEqual(mock_logger.mock_calls, expected_calls)
-
-  @mock.patch('google_compute_engine.clock_skew.clock_skew_daemon.subprocess.check_call')
-  def testHandleClockSyncError(self, mock_call):
-    command = ['/sbin/hwclock', '--hctosys']
-    mock_sync = mock.create_autospec(clock_skew_daemon.ClockSkewDaemon)
-    mock_logger = mock.Mock()
-    mock_sync.logger = mock_logger
-    mock_call.side_effect = subprocess.CalledProcessError(1, 'Test')
-
-    clock_skew_daemon.ClockSkewDaemon.HandleClockSync(mock_sync, 'Response')
-    mock_call.assert_called_once_with(command)
     expected_calls = [
-        mock.call.info(mock.ANY, 'Response'),
-        mock.call.warning(mock.ANY),
+        mock.call.HandleClockSync(mock_logger),
     ]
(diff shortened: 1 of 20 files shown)


google_compute_engine/distro/tests/helpers_test.py
```diff
@@ -64,3 +64,84 @@ def testCallDhclient(self, mock_call, mock_exists):
...
     self.assertEqual(mocks.mock_calls, expected_calls)
+
+  @mock.patch('google_compute_engine.distro.helpers.subprocess.check_call')
+  def testCallHwclock(self, mock_call):
+    command = ['/sbin/hwclock', '--hctosys']
+    mock_logger = mock.Mock()
+
+    helpers.CallHwclock(mock_logger)
+    mock_call.assert_called_once_with(command)
+    expected_calls = [
```
illfelder: This fits on one line in under 80 characters; no need to split it up.

This also applies to the `expected_calls` below.

google_compute_engine/clock_skew/tests/clock_skew_daemon_test.py
```diff
@@ -71,36 +71,22 @@ def testClockSkewDaemonError(self, mock_lock, mock_logger, mock_watcher):
...
+  @mock.patch('google_compute_engine.clock_skew.clock_skew_daemon.distro_utils')
+  def testHandleClockSync(self, mock_distro_utils):
     mock_sync = mock.create_autospec(clock_skew_daemon.ClockSkewDaemon)
     mock_logger = mock.Mock()
     mock_sync.logger = mock_logger
+    mock_sync.distro_utils = mock_distro_utils
 
     clock_skew_daemon.ClockSkewDaemon.HandleClockSync(mock_sync, 'Response')
-    mock_call.assert_called_once_with(command)
     expected_calls = [
```
illfelder: You can move this onto one line (and below).

google_compute_engine/distro/utils.py
```diff
@@ -37,3 +39,13 @@ def EnableNetworkInterfaces(
...
+
+  def HandleClockSync(self, logger):
+    """Sync clock.
+
+    Called when clock drift token changes.
+
+    Args:
+      logger: logger object, used to write to SysLog and serial port.
+    """
+    helpers.CallHwclock(logger)
```
illfelder: /cc @ryanwe 

This file was intended as an interface. The function definitions belong here, but the implementation exists in the distro specific directories. Using this paradigm, introducing a new function definition requires an update to each distro utils file. Not sure if we should change the design, but we should give it some thought.

helen-fornazier:
> This file was intended as an interface. The function definitions belong here, but the implementation exists in the distro specific directories. Using this paradigm, introducing a new function definition requires an update to each distro utils file. Not sure if we should change the design, but we should give it some thought.

This was my first approach to the implementation (I have another branch with it [1]), but implementing in the parent folder avoids several repetitions of the code.
Let me know which implementation you prefer and I'll update the PR accordingly

[1] https://github.com/helen-fornazier/compute-image-packages/commit/bf79b7105c6dc350e67e46892ecccfa53af4c7c0

illfelder:
>> This file was intended as an interface. The function definitions belong here, but the implementation exists in the distro specific directories. Using this paradigm, introducing a new function definition requires an update to each distro utils file. Not sure if we should change the design, but we should give it some thought.
>
> This was my first approach to the implementation (I have another branch with it [1]), but implementing in the parent folder avoids several repetitions of the code.
Let me know which implementation you prefer and I'll update the PR accordingly
>
> [1] https://github.com/helen-fornazier/compute-image-packages/commit/bf79b7105c6dc350e67e46892ecccfa53af4c7c0

We discussed this internally. We think we should keep `utils.py` an interface where each function throws an unimplemented exception instead of doing nothing. This will make it more obvious when new functionality is introduced that isn't getting overridden by one of the distros. It's fine if each distro implements the functionality by calling the helpers function, though we may way to split up helpers at some point if it becomes unwieldy.

Let me know if you have any questions/concerns with this approach.

helen-fornazier:
@illfelder could you review this last version please?

google_compute_engine/distro/utils.py
```diff
@@ -37,3 +37,13 @@ def EnableNetworkInterfaces(
       dhclient_script: string, the path to a dhclient script used by dhclient.
     """
     pass
+
+  def HandleClockSync(self, logger):
+    """Sync the software clock with the hypervisor clock.
```
illfelder: Can you update the other distro util files to have the same description? Also, please remove the "Called when..." line below.

helen-fornazier: Done

Issue #603, opened 2018-05-10
google-network-daemon / google-ip-forwarding in systemd


mills95:
The recent refactoring (2018) has caused some issues with knowing when the system has obtained the metadata and parsed the forwarding IPs. When google-ip-forwarding was a standalone daemon, we could wait for it to complete before starting processes that depend on having those IPs.  However, now that it is dynamically spawned, this is no longer possible.  Is there an option to force google-ip-forwarding to run before google-network-daemon returns so we can wait on that?
Right now, we have to put in a time delay which is both problematic and wasteful; we don't want to miss the completion, but we also want the server to start up as quickly as possible. But we have to have something that works if our instance is migrated.

illfelder:
Thanks for the report! We found an issue where we introduced a ~2 minute delay when setting IP forwarding routes. We are actively working on a fix.

illfelder:
The fix is merged, release, and included in the latest Debian, CentOS, and RHEL images.

PR #605, opened 2018-05-10
Prevent delay in configuring IP forwarding routes.

illfelder:
Add support for enabling network interfaces if they appear while the VM
is running.

### DIFF
--- a/google_compute_engine/networking/network_daemon.py
+++ b/google_compute_engine/networking/network_daemon.py
@@ -58,26 +58,17 @@ class NetworkDaemon(object):
     facility = logging.handlers.SysLogHandler.LOG_DAEMON
     self.logger = logger.Logger(
         name='google-networking', debug=debug, facility=facility)
-    self.network_utils = network_utils.NetworkUtils(logger=self.logger)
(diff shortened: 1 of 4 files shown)


illfelder:
Fixes #603

google_compute_engine/networking/network_daemon.py
```diff
@@ -58,26 +58,19 @@ def __init__(
...
-    self.network_utils = network_utils.NetworkUtils(logger=self.logger)
-    self.watcher = metadata_watcher.MetadataWatcher(logger=self.logger)
-
+    self.dhclient_script = dhclient_script
+    self.dhcp_command = dhcp_command
     self.ip_aliases = ip_aliases
+    self.ip_forwarding_enabled = ip_forwarding_enabled
+    self.network_setup_enabled = network_setup_enabled
     self.target_instance_ips = target_instance_ips
+    self.debug = debug
```
ryanwe: Some of these statements are no longer needed.

illfelder: Thanks!

Issue #602, opened 2018-05-10
GetForwardedIps issues on RHEL7


mills95:
When GetForwardedIps runs _RunIpRoute (effective command: ip route ls table local type local scope host dev eth0 proto 66), the result is 0 or more lines. If there is one or more forwarded ips, they are returned one per line but have "local " prepended to them.  This causes ParseForwardedIps to fail.

That may not be the case on other systems, but a simple solution works for all of them.  Before returning from GetForwardedIps, simply do a search & replace:
`result = re.sub(r'local\s((\d{1,3}\.){3}\d{1,3}(/\d{1,2})?)\s?\n', r'\1\n',result)`

then return as normal.  If there is a "local " tag, it gets removed, if not, nothing changes.
This helped me stop thousands of warnings from filling up /var/log/messages with:
  google-ip-forwarding: WARNING Could not parse IP address: "local".

mills95:
I can now confirm that this also occurs on Centos7 after the recent upgrade from iproute-3.10x to iproute-4.11x.  It also occurs on Ubuntu 17. Couldn't find the exact point in the change list where ip route changed it's return value, but seems most likely to be in the iproute-4.x tree, and RedHat upgraded 
 recently (in Mar as far as I can tell from the logs) which is slowly percolating as Centos just got it in the last week.

illfelder:
Hi @mills95, thanks for the report! Can you provide a sample of the output from `_RunIpRoute` command that has `local` prepended?

mills95:
The default command run via subprocess is "ip route ls table local type local scope host dev eth0 proto 66", but that may be slightly different based on the specific instance.
The output will looks like (values changed):

local 35.185.160.XX 
local 35.185.160.YY 

Note the space at the end.

Basically, just run "ip route ls table local", if the output starts with the type (local, broadcast, etc), then you are running the updated iproute.  The more qualifiers you add, the closer you get to just an IP address, but with the updated version you never completely get there unlike the old version where you would eventually get a bare IP.

illfelder:
This issue is fixed in the latest release.

jfreeland:
I'm still seeing this on COS 10895.277.0.

brendan-rius:
I have the same error message in the syslog of one of my GKE nodes running 1.13.7-gke.8, which I think is cos-u-73-11647-182-0 according to https://cloud.google.com/kubernetes-engine/docs/release-notes

[...]

PR #608, opened 2018-05-13
Prevent log spam from the IP forwarding daemon.

illfelder:
Forwarded IPs are returned with a local prefix. The agent splits on the
space and tries to add a route for the IP "local" which generates log
spam. We should strip out the local prefix to prevent this issue.


illfelder:
Fixes #602.

PR #616, opened 2018-05-21
Lnx distro py3

rjschwei:
Handle distro detection in Python 3.6 and greater.


rjschwei:
Obviously I am incapable of doing this correctly

PR #617, opened 2018-05-21
- Properly handle distro detection for Python 3.6 and greater

rjschwei:
+ The present implementation fails on relative import of "distro" and
    "distro" does not have an implementation of "linux_distribution" which
    is called in any Python version. This commit addresses both problems,
    changing the import to absolute and implementing "linux_distribution"
    to return values based on "/etc/os-release"

### DIFF
--- a/google_compute_engine/compat.py
+++ b/google_compute_engine/compat.py
@@ -20,7 +20,7 @@ import subprocess
 import sys
 
 if sys.version_info >= (3, 6):
-  import distro
+  import google_compute_engine.distro as distro
 else:
   import platform as distro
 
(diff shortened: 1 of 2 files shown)


rjschwei:
Third time's a charm, just need to pick the right target branch.

zmarano:
This one :)
https://software.opensuse.org/package/python3-distro

rjschwei:
We package the distro implementation and I can certainly add a dependency to our package. However, given the naming we may still be in trouble especially since upstream Python more or less changes relative import rules as they see fit at any given time. Also depending on user manipulation of PYTHONPATH.

Since a directory with the name "distro" exists, relative to compat.py it is more or less easy to trick Python to load google_compute_engine.distro which then would load the empty init file. This "confusion" is also what led me to implement linux_distribution() in __init__.py.

If the intention is to use the distro module I would strongly recommend that google_compute_engine/distro get renamed

PR #619, opened 2018-05-31
Rename distro directory to distro_lib.

illfelder:
This prevents conflicts with the Python 3 distro module.


illfelder:
/cc @rjschwei 

Fixes #617

PR #622, opened 2018-06-10
class IpForwardingUtils implemented in distro_lib

helen-fornazier:
The old implementation of IpForwardingUtils uses ip route command, but
this is not suported in FreeBSD. This commit moves the implementation to
distro_lib and transforms IpForwardingUtils into an interface class,
allowing each distro to chose which implementation to use.

class IpForwardingUtilsIprouteTest:
	contains the old implemetation using ip route command
class IpForwardingUtilsIfconfig:
	contains an implementation using ifconfig command


illfelder:
As currently written, anyone taking a dependency on a method in the IpForwardingUtils class will be broken by this change. If you are looking to move the functionality to the distros library, the utils file should be inheriting the functionality from the distro utils.

helen-fornazier:
Right, just a question: are those python libs only used by these daemons? Or do you expect other people to reuse the utils to another purpose?

I am asking this because I could completely delete file `google_compute_engine/distro_lib/ip_forwarding_utils.py` and move the abstract class to distro_libs
Like this, daemons can't have a dependency in the IpForwardingUtils class directly, as they are supposed access distro_lib/ only thought `from google_compute_engine.compat import distro_utils`, so essentially, file `google_compute_engine/distro_lib/ip_forwarding_utils.py` is a helper file.

I updated the PR with a new commit to show you what I mean. Please let me know what you think.

illfelder:
> Right, just a question: are those python libs only used by these daemons? Or do you expect other people to reuse the utils to another purpose?

No. Historically we've seen many reports from users who have install the guest environment via pip to a virtual env. Users write their own code taking a dependence on our libraries. If it's a private function, we make no guarantees they won't be changed or removed, but public functions should maintain backwards compatibility.

helen-fornazier:
@illfelder could you take a look please in this last modification of this PR to check if the structure is ok? Then I'll proceed to implement the missing mock tests.

google_compute_engine/distro_lib/ip_forwarding_utils.py
```diff
@@ -0,0 +1,303 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Utilities for configuring IP address forwarding."""
+
+import re
+import subprocess
+try:
+  # The following libs are required only for class IpForwardingUtilsIfconfig
```
illfelder: Please replace this with:
`# The following modules are required by IpForwardingUtilsIfconfig.`

google_compute_engine/networking/ip_forwarding/ip_forwarding_utils.py
```diff
@@ -15,11 +15,7 @@
...
 
-import re
-import subprocess
-
-IP_REGEX = re.compile(r'\A(\d{1,3}\.){3}\d{1,3}\Z')
-IP_ALIAS_REGEX = re.compile(r'\A(\d{1,3}\.){3}\d{1,3}/\d{1,2}\Z')
+from google_compute_engine.compat import distro_utils
 
 
 class IpForwardingUtils(object):
```
illfelder: I think you can replace this class with the following

```
class IpForwardingUtils(distro_utils.IpForwardingUtils):
  """Deprecated. Overridden for backwards compatibility."""
  pass
```

helen-fornazier: To do this I had to add in each distro specific file the following:

    +from google_compute_engine.distro_lib.ip_forwarding_utils import \
    +    IpForwardingUtilsIproute as IpForwardingUtils

And I removed the function `def IpForwardingUtils` from `class Utils`, please let me know if this is ok or if you had another idea in mind.

google_compute_engine/distro_lib/ip_forwarding_utils.py
helen-fornazier: done

google_compute_engine/distro_lib/ip_forwarding_utils.py
```diff
@@ -0,0 +1,303 @@
...
+except ImportError:
+  netifaces = None
+  netaddr = None
+
+
+IP_REGEX = re.compile(r'\A(\d{1,3}\.){3}\d{1,3}\Z')
+IP_ALIAS_REGEX = re.compile(r'\A(\d{1,3}\.){3}\d{1,3}/\d{1,2}\Z')
+
+
+class IpForwardingUtils(object):
```
wrigri: So, I wonder if we could make this a bit clearer. A couple thoughts:
- Could we rename "IpForwardingUtils" to "IpForwardingUtilsBase" or something and then have IpForwardingUtils be a dummy class of some sort that will return the correct class based on a constant that we can override or something? (I admit that I'm not 100% sure how to make that work, so if it's not feasible, then no worries.)

A different idea:
In the individual utils.py files, instead of doing the override as part of the import, could we instead do something like this:

from google_compute_engine.distro_utils import ip_forwarding_utils
ip_forwarding_utils.IpForwardingUtils = ip_forwarding_utils.IpForwardingUtilsIfconfig

I believe that the result would be similar to the imports that we are doing now, but it seems like it would be a little clearer about what's going on.

helen-fornazier: What do you think if we instead use a function in the utils.py as the previous commit [1], and use the `__new__` function in the original class `google_compute_engine.networking.ip_forwarding.IpForwardingUtils` which will return an instance created by utils.py ?
This would keep the utils.py with the same pattern (overriding functions).


[1] https://github.com/GoogleCloudPlatform/compute-image-packages/pull/622/commits/1a9dfb323793c11ac7f3b15ec687d665c262e4eb#diff-ffd46ac80f68d2b94cbe7060700b93b5R49

gut: Hi. I would also force the `IpForwardingUtils` to be an abstract class (by using the abs module) instead of only empty implementation. That would improve maintainability when extending new classes.

wrigri: To the extent that I understand what you are saying, it seems reasonable. (I haven't used __new__ extensively, so it's hard for me to be 100% confident in that.)

I do agree that setting it up as an abstract class would help in the long run.

Mainly, though, I think the most important thing is just making it so that it's clear what's going on, and the override as part of the import wasn't doing that.

helen-fornazier:
@illfelder @wrigri Could you please check this last commit using the __new__ function?

google_compute_engine/distro_lib/ip_forwarding_utils.py
```diff
@@ -0,0 +1,303 @@
+#!/usr/bin/python
+# Copyright 2016 Google Inc. All Rights Reserved.
```
illfelder: Please update the copyright.

google_compute_engine/distro_lib/ip_forwarding_utils.py
```diff
@@ -0,0 +1,303 @@
...
+      ips = netifaces.ifaddresses(interface)
+      ips = ips[netifaces.AF_INET]
+    except (ValueError, IndexError):
+      return []
+    forwarded_ips = []
+    for ip in ips:
+      if ip['addr'] != interface_ip:
+        forwarded_ips.append(
+            '%s/%d' % (ip['addr'],
+            netaddr.IPAddress(ip['netmask']).netmask_bits()))
```
illfelder: Can you pull this into a variable to make this line easier to understand?

helen-fornazier:
@illfelder Sorry the delay to update this. Could you please check the last commit? I changed the name of the base class in distro_lib here https://github.com/GoogleCloudPlatform/compute-image-packages/pull/622/commits/d505f5a27f6e136d4b8a59bfc38ad4626b552bd6#diff-c26ce3dffc0246c687d323f46d25913cL33 to not confuse with the original class from `google_compute_engine/networking/ip_forwarding/ip_forwarding.py`

I was also thinking in changing the name of this function at https://github.com/GoogleCloudPlatform/compute-image-packages/pull/622/files#diff-5fe45074a645185b58e3d918a8bfd0ffR45 to not confuse with the name of the class, what do you think?



Tests by importing the lib directly:
```
Python 2.7.13 (default, Nov 24 2017, 17:33:09)
[GCC 6.3.0 20170516] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from google_compute_engine.networking.ip_forwarding import ip_forwarding_utils
>>> import logging
>>> ipf = ip_forwarding_utils.IpForwardingUtils(logging, '66')
>>> ipf.GetForwardedIps('eth0')
[u'10.128.0.12/30', u'10.128.0.22']
>>> ipf.AddForwardedIp('10.128.0.02/31', 'eth0')
>>> ipf.GetForwardedIps('eth0')
[u'10.128.0.2/31', u'10.128.0.12/30', u'10.128.0.22']
>>> ipf.RemoveForwardedIp('10.128.0.12/30', 'eth0')
>>> ipf.GetForwardedIps('eth0')
[u'10.128.0.2/31', u'10.128.0.22']
>>> ipf.ParseForwardedIps(['10.128.4.12', '123.123.123.123/24', 'adsf', '192.168.0.15'])
WARNING:root:Could not parse IP address: "adsf".
['10.128.4.12', '123.123.123.123/24', '192.168.0.15']
```

illfelder:
Apologies, I thought I already reviewed, approved and merged this!

PR #624, opened 2018-07-03
OSLogin partially fails when username is greater than 32 characters

jrw972:
A user with a username greater than 32 characters (name and email with
special characters converted to _) logged into an instance with
OSLogin enabled.  The user received "No user exists for uid XXXX"
messages when attempting to run programs.  This change truncates the
username to 32 characters for utilities that take a username.

### DIFF
--- a/google_compute_engine/accounts/accounts_utils.py
+++ b/google_compute_engine/accounts/accounts_utils.py
@@ -26,7 +26,7 @@ import tempfile
 from google_compute_engine import constants
 from google_compute_engine import file_utils
 
-USER_REGEX = re.compile(r'\A[A-Za-z0-9._][A-Za-z0-9._-]*\Z')
+USER_REGEX = re.compile(r'\A[A-Za-z0-9._][A-Za-z0-9._-]{0,31}\Z')
 DEFAULT_GROUPADD_CMD = 'groupadd {group}'
 DEFAULT_USERADD_CMD = 'useradd -m -s /bin/bash -p * {user}'
 DEFAULT_USERDEL_CMD = 'userdel -r {user}'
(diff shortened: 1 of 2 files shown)


illfelder:
Hi, thanks for the contribution. Most Linux systems do not allow usernames that are longer than 32 characters and OS Login does not permit users to set their username beyond that length. When you attempt to SSH into an instance, you should be using the username in your login profile (the one that is provided when you [import an SSH key](https://developers.google.com/apis-explorer/#p/oslogin/v1/oslogin.users.importSshPublicKey)). It's plausible that your username is not the same as your sanitized email address, so truncating at 32 characters is not a behavior I believe we should encourage.

As a note, the changes you made in your PR are modifying the accounts daemon (rather than OS Login). The accounts daemon should also validate that usernames are not longer than 32 characters, but that should be done as an update to the validation logic:
```
USER_REGEX = re.compile(r'\A[A-Za-z0-9._][A-Za-z0-9._-]*\Z')
```

illfelder:
Hi, thanks for updating the PR - the change looks good to me (though changes should diff against our development branch rather than master).

I looked through the code that handles SSH from the GCP console. The tool that allows you to log in will automatically derive a username for you based on your email. The username generation logic has a maximum length of 32 characters. If you are running into a length limit while using the tool, it's likely that you have a custom username set.

Regarding the CLA issue, I'm not seeing anything associated with the user `jrw972`. Can you visit https://cla.developers.google.com/clas to double check you've signed the CLA for that user?

jrw972:
Can you check the CLA again for wilsonj@objectcomputing.com?

illfelder:
Still don't see anything yet unfortunately...

jrw972:
I think the CLA is now signed.

PR #632, opened 2018-07-26
Enable IPv6 on interfaces based on metadata key.

illfelder:
Support for FreeBSD and SLES images is still needed.

### DIFF
--- a/packages/python-google-compute-engine/google_compute_engine/distro_lib/helpers.py
+++ b/packages/python-google-compute-engine/google_compute_engine/distro_lib/helpers.py
@@ -42,6 +42,27 @@ def CallDhclient(
     logger.warning('Could not enable interfaces %s.', interfaces)
 
 
+def CallDhclientIpv6(interfaces, logger, dhclient_script=None):
+  """Configure the network interfaces for IPv6 using dhclient.
+
+  Args:
+    interface: string, the output device names for enabling IPv6.
+    logger: logger object, used to write to SysLog and serial port.
+    dhclient_script: string, the path to a dhclient script used by dhclient.
+  """
+  logger.info('Enabling IPv6 on the Ethernet interfaces %s.', interfaces)
+
+  dhclient_command = ['dhclient']
+
+  if dhclient_script and os.path.exists(dhclient_script):
+    dhclient_command += ['-sf', dhclient_script]
+
+  try:
+    subprocess.check_call(dhclient_command + ['-1', '-6', '-v'] + interfaces)
+  except subprocess.CalledProcessError:
+    logger.warning('Could not enable IPv6 on interface %s.', interfaces)
+
+
 def CallHwclock(logger):
   """Sync clock using hwclock.
 
--- a/packages/python-google-compute-engine/google_compute_engine/distro_lib/helpers.py
+++ b/packages/python-google-compute-engine/google_compute_engine/distro_lib/helpers.py
@@ -52,13 +52,15 @@ def CallDhclientIpv6(interfaces, logger, dhclient_script=None):
   """
   logger.info('Enabling IPv6 on the Ethernet interfaces %s.', interfaces)
 
+  timeout_command = ['timeout', '5']
   dhclient_command = ['dhclient']
 
   if dhclient_script and os.path.exists(dhclient_script):
     dhclient_command += ['-sf', dhclient_script]
 
   try:
-    subprocess.check_call(dhclient_command + ['-1', '-6', '-v'] + interfaces)
+    subprocess.check_call(
+        timeout_command + dhclient_command + ['-1', '-6', '-v'] + interfaces)
   except subprocess.CalledProcessError:
     logger.warning('Could not enable IPv6 on interface %s.', interfaces)
 
--- a/packages/python-google-compute-engine/google_compute_engine/networking/network_daemon.py
+++ b/packages/python-google-compute-engine/google_compute_engine/networking/network_daemon.py
@@ -90,6 +90,9 @@ class NetworkDaemon(object):
     network_interfaces = self._ExtractInterfaceMetadata(result)
 
     if self.network_setup_enabled:
+      default_interface = network_interfaces[0]
+      if default_interface.ipv6:
+        self.network_setup.EnableIpv6([default_interface.name])
       self.network_setup.EnableNetworkInterfaces(
           [interface.name for interface in network_interfaces[1:]])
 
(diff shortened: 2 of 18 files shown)


helen-fornazier:
just to confirm, the interface should use the stateful DHCPv6 instead of the stateless autoconfiguration from the router advertisements correct?

Also, I would like to test this in GCE, but the only doc I found was related to load-balancing, is there an easier way to enable ipv6 to a GCE instance?

helen-fornazier:
@illfelder 

Also, sles have dhclient with the same options supported, but it requires package `dhcp-client` and sles 11 doesn't came with it by default.
In the packaging folder I see build scripts for debian, el6 and el7, but I don't see any scripts for sles, is it the same used for el6 and el7? I just want to check with you if the file `specs/python-google-compute-engine.spec` is the best place to add the requirement to install `dhcp-client` for sles

Thanks

zmarano:
The SLES (and OpenSUSE) packages are maintained by SUSE here: https://build.opensuse.org/package/show/Cloud:Tools/google-compute-engine

gut:
> I just want to check with you if the file `specs/python-google-compute-engine.spec` is the best place to add the requirement to install `dhcp-client` for sles

Hi @helen-fornazier . I submitted a bug report to SLES for having the dhcp-client installed on SLES11 and I've got a useful information for you:

https://bugzilla.suse.com/show_bug.cgi?id=1106323#c3

Check if it'll indeed work properly on SLES 12 and beyond as dhcp-client will not be controlling the network card. You should be using wicked instead (see link above).

packages/python-google-compute-engine/google_compute_engine/distro_lib/helpers.py
```diff
@@ -42,6 +42,27 @@ def CallDhclient(
...
+  """
+  logger.info('Enabling IPv6 on the Ethernet interfaces %s.', interfaces)
+
+  dhclient_command = ['dhclient']
+
+  if dhclient_script and os.path.exists(dhclient_script):
+    dhclient_command += ['-sf', dhclient_script]
+
+  try:
+    subprocess.check_call(dhclient_command + ['-1', '-6', '-v'] + interfaces)
```
hopkiw: why these args?

packages/python-google-compute-engine/google_compute_engine/networking/network_setup/network_setup.py
```diff
@@ -43,6 +44,22 @@ def __init__(self, dhclient_script=None, dhcp_command=None, debug=False):
...
+  def EnableIpv6(self, interfaces):
+    """Enable IPv6 on the list of network interfaces.
+
+    Args:
+      interfaces: list of string, the output device names for enabling IPv6.
+    """
+    if not interfaces or set(interfaces) == self.ipv6_interfaces:
+      return
+
+    self.logger.info('Enabling IPv6 on Ethernet interface: %s.', interfaces)
```
hopkiw: three references to this var, this is the only one not using the set() version. maybe interfaces = set(interfaces) as the first statement in this method?

packages/python-google-compute-engine/google_compute_engine/networking/network_daemon.py
```diff
@@ -119,7 +122,9 @@ def _ExtractInterfaceMetadata(self, metadata):
         if self.target_instance_ips:
           ip_addresses.extend(network_interface.get('targetInstanceIps', []))
         interfaces.append(NetworkDaemon.NetworkInterface(
-            interface, ip_addresses, network_interface.get('ip', [])))
+            interface, forwarded_ips=ip_addresses,
+            ip=network_interface.get('ip', None),
```
hopkiw: why change [] to None ?

packages/python-google-compute-engine/google_compute_engine/networking/network_daemon.py
```diff
@@ -119,7 +122,9 @@ def _ExtractInterfaceMetadata(self, metadata):
...
+            ipv6='dhcpv6Refresh' in network_interface.keys()))
```
hopkiw: this is the metadata key check? should it be case sensitive like this?

packages/python-google-compute-engine/google_compute_engine/networking/network_daemon.py
illfelder: It's not changing. This is saying retrieve `ip` and if it not set, return None. The `NetworkInterface` object has None as the default rather than an empty list.

packages/python-google-compute-engine/google_compute_engine/networking/network_daemon.py
illfelder: Yes, since it's defined by internal systems.

packages/python-google-compute-engine/google_compute_engine/networking/network_setup/network_setup.py
illfelder: Done. Also updated the function below for consistency.

packages/python-google-compute-engine/google_compute_engine/distro_lib/helpers.py
hopkiw: OK, looked into the issue driving this change. Notes:
`-6` is obviously needed. 
`-v` is quite noisy, i think since we are wrapping these system commands we should drop it.
`-1` has a behavior change between IPv4 and IPv6, where in the latter case it forces the refresh to limit itself to the timeout setting from dhclient.conf. However for some systems (e.g. ubuntu 1804), this is set to 5 minutes. In any DHCP failure case, that will hang the entire network daemon for five minutes.
Recommend running this in the background to protect against that case and possibly adding some checks to see if we have any running before doing this.

hopkiw: ah and `-x` from our normal dhclient calls is not needed, that only kills established dhclient daemons

acornejo: the timeout can be easily enforced by hand by prepending the command with the 'timeout' comamnd, i.e.:

timeout dhclient -1 -6 eth0

The timeout command is part of the coreutils, so its available in all fedora/debian/etc based systems by default.

illfelder: Another alternative is to fork and run the command as a backgrounded process. I'm not sure which approach is better. @zmarano @hopkiw any preferences?

acornejo: FWIW if you choose to background the process some care may be needed to make sure you don't fork too many such clients in cases where the interface goes up and down. you could try to kill previous processes to make sure you only run one at a time, but then you could worry about how to kill the "correct" dhclient since the user may be running instances of their own. For that you could write a pid file, etc..

You could also choose to ignore these nuances and hope we simply dont run into these corner cases. Not sure what level of paranoia is warranted here.

hopkiw: for me personally a simple safeguard like timeout or setting the timeout argument to subprocess.check_call() would be reasonable. as i understand it, this 5m hang condition would be reached any time dhcp fails for a secondary interface, and the attempts are triggered by a metadata key.

PR #635, opened 2018-08-01
Fix python packaging doc URL on README.md

gut:


illfelder:
I updated the GitHub branch to diff against development, but it looks like you started off of master. Please rebase your branch against development and I'll merge this in.

gut:
your assumption was correct. I fixed it. Thanks

PR #637, opened 2018-08-18
Remove gsutil dependency

gut:


google_compute_engine/metadata_scripts/script_retriever.py
```diff
@@ -111,7 +127,10 @@ def _DownloadScript(self, url, dest_dir):
     # Check for the preferred Google Storage URL format:
     # gs://<bucket>/<object>
     if url.startswith(r'gs://'):
-      return self._DownloadGsUrl(url, dest_dir)
+      # convert it to a regular URL
+      bucket_and_object = url[len(r'gs://'):]
+      url = 'https://storage.googleapis.com/%s' % (bucket_and_object)
```
illfelder: Please remove the extra parens around the variable. In general, use parens sparingly (if they are not needed, don't add them).

google_compute_engine/metadata_scripts/script_retriever.py
```diff
@@ -111,7 +127,10 @@ def _DownloadScript(self, url, dest_dir):
     # Check for the preferred Google Storage URL format:
     # gs://<bucket>/<object>
     if url.startswith(r'gs://'):
-      return self._DownloadGsUrl(url, dest_dir)
+      # convert it to a regular URL
```
illfelder: `s/convert it to a regular URL/Convert the string into a standard URL./`

google_compute_engine/metadata_scripts/script_retriever.py
```diff
@@ -111,7 +127,10 @@ def _DownloadScript(self, url, dest_dir):
...
+      bucket_and_object = url[len(r'gs://'):]
```
illfelder: `url =  re.sub('^gs://', 'https://storage.googleapis.com/', url)`

google_compute_engine/metadata_scripts/script_retriever.py
```diff
@@ -50,29 +55,40 @@ def _DownloadGsUrl(self, url, dest_dir):
...
 
-    self.logger.info('Downloading url from %s to %s using gsutil.', url, dest)
+    self.logger.info(
+        'Downloading url from %s to %s using authentication token.',
+        url, dest)
     try:
-      subprocess.check_call(['gsutil', 'cp', url, dest])
+      if not self._token:
+        tok_url = '%(metadata)s/instance/service-accounts/default/token' % {
+            'metadata': 'http://metadata.google.internal/computeMetadata/v1',
```
illfelder: I don't follow. What's the benefit of this dictionary over defining the constant?

google_compute_engine/metadata_scripts/script_retriever.py
```diff
@@ -50,29 +55,40 @@ def _DownloadGsUrl(self, url, dest_dir):
...
+        }
+        request = urlrequest.Request(tok_url)
+        request.add_unredirected_header('Metadata-Flavor', 'Google')
```
illfelder: You're requesting metadata from inside the Linux guest environment. You can (and should) use the metadata_watcher library which is what we use elsewhere in this code.

google_compute_engine/metadata_scripts/script_retriever.py
```diff
@@ -50,29 +55,40 @@ def _DownloadGsUrl(self, url, dest_dir):
...
+        # converts the stringified dictionary of the response to a dictionary
```
illfelder: All comments are sentences - pease capitalize the first letter and end in a period.

google_compute_engine/metadata_scripts/tests/script_retriever_test.py
```diff
@@ -33,47 +34,66 @@ def setUp(self):
...
+    mock_request.return_value = mock_request
+    mock_urlopen.return_value = mock_urlopen
+    mock_urlopen.read.return_value = "{'token_type':'foo', 'access_token':'bar'}"
+
+    # Throw exception only on 2nd call.
+    mock_request.side_effect = [mock_request, Exception('Error.')]
+
+    self.assertIsNone(self.retriever._DownloadAuthUrl(auth_url, self.dest_dir))
+
+    self.assertEqual(mock_request.mock_calls,
```
illfelder: If a line exceeds 80 characters, always line wrap from the first open paren.

google_compute_engine/metadata_scripts/script_retriever.py
gut: line breaking. Or should I just place the whole string (which is awesome when searching code) and add a flake8 warning dismiss?

google_compute_engine/metadata_scripts/script_retriever.py
gut: I was unaware of that library. Thanks

google_compute_engine/metadata_scripts/script_retriever.py
illfelder: If you have a long line like this, it's preferred to break it up by relying on string concatenation:

```
tok_url =(
    'http://metadata.google.internal/computeMetadata/v1/'
    'instance/service-accounts/default/token')
```

Having said that, this shouldn't be necessary once you move to using the metadata watcher.

gut:
@illfelder Thanks max. I hope to have fulfilled your comments accordingly.

google_compute_engine/metadata_scripts/script_retriever.py
```diff
@@ -50,29 +55,35 @@ def _DownloadGsUrl(self, url, dest_dir):
...
+    self.logger.info(
+        'Downloading url from %s to %s using authentication token.',
+        url, dest)
     try:
-      subprocess.check_call(['gsutil', 'cp', url, dest])
+      if not self._token:
+        tok_metadata = 'instance/service-accounts/default/token'
+        response = self.watcher.GetMetadata(tok_metadata, recursive=False)
+        self._token = '%s %s' % (
+            response[u'token_type'], response[u'access_token']
```
illfelder:
```
self._token = '%s %s' % (
    response.get('token_type', ''), response.get('access_token', ''))
````

File: google_compute_engine/metadata_scripts/script_retriever.py [hunk 9]
```diff
@@ -50,29 +55,35 @@ def _DownloadGsUrl(self, url, dest_dir):
...
+      content = urlrequest.urlopen(request).read()
+
+      with open(dest, 'w') as f:
+        f.write(content)
+
       return dest
-    except subprocess.CalledProcessError as e:
-      self.logger.warning(
-          'Could not download %s using gsutil. %s.', url, str(e))
     except Exception as e:
```
illfelder: This try except block seems fairly large. You should scope the block to areas that throw exceptions (potentially have two), and then handle the exceptions you expect. I don't think you should need to catch a generic Exception.

google_compute_engine/metadata_scripts/tests/script_retriever_test.py
```diff
@@ -33,47 +35,64 @@ def setUp(self):
...
+    handle = mock_open()
+    handle.write.assert_called_once_with(urlopen_read)
 
-  @mock.patch('google_compute_engine.metadata_scripts.script_retriever.subprocess.check_call')
   @mock.patch('google_compute_engine.metadata_scripts.script_retriever.tempfile.NamedTemporaryFile')
-  def testDownloadGsUrlException(self, mock_tempfile, mock_call):
-    gs_url = 'gs://fake/url'
+  @mock.patch('google_compute_engine.metadata_scripts.script_retriever.urlrequest.Request')
+  @mock.patch('google_compute_engine.metadata_watcher.MetadataWatcher.GetMetadata')
+  def testDownloadAuthUrlExceptionAndToken(self, mock_GetMetadata,
```
illfelder: Please line wrap at the first open paren.

google_compute_engine/metadata_scripts/tests/script_retriever_test.py
```diff
@@ -33,47 +35,64 @@ def setUp(self):
...
+    self.retriever._token = None
+
+    mock_GetMetadata.return_value = {'token_type':'foo', 'access_token':'bar'}
+    mock_request.return_value = mock_request
+    mock_request.side_effect = Exception('Error.')
+
+    self.assertIsNone(self.retriever._DownloadAuthUrl(auth_url, self.dest_dir))
+
+    mock_GetMetadata.return_value = mock_GetMetadata
+    # GetMetadata includes a prefix, so remove it
```
illfelder: Please end comments with periods.

google_compute_engine/metadata_scripts/tests/script_retriever_test.py
```diff
@@ -171,25 +191,28 @@ def testDownloadScript(self):
       download_urls.extend(urls)
       download_gs_urls.update(gs_urls)
 
+    # all gs:// urls uses authentication to be downloaded
```
illfelder: Please replace this comment with:
```
# All Google Storage URLs are downloaded with an authentication token.
```

gut:
hum... thanks to new metadata-scripts tests without scopes I can see that this fix is not ideal: if a `gs://` script-url is given, then it might be possible that it is a public URL and there is no authentication token that can be used. Therefore, if authenticated download fails, try unauthenticated. 

I'll update this shortly.

google_compute_engine/metadata_scripts/script_retriever.py
```diff
@@ -40,8 +45,9 @@ def __init__(self, logger, script_type):
     self.script_type = script_type
     self.watcher = metadata_watcher.MetadataWatcher(logger=self.logger)
 
-  def _DownloadGsUrl(self, url, dest_dir):
-    """Download a Google Storage URL using gsutil.
+  def _DownloadAuthUrl(self, url, dest_dir):
+    """Download a Google Storage URL using an authentication token. If token
```
illfelder:
```
"""Download a Google Storage URL using an authentication token.

If the token cannot be fetched, fallback to an unauthenticated download.

Args:
```

google_compute_engine/metadata_scripts/script_retriever.py
```diff
@@ -50,29 +56,38 @@ def _DownloadGsUrl(self, url, dest_dir):
...
-    self.logger.info('Downloading url from %s to %s using gsutil.', url, dest)
+    self.logger.info(
+        'Downloading url from %s to %s using authentication token.', url, dest)
+
+    if not self.token:
+      response = self.watcher.GetMetadata(
+          self.token_metadata_key, recursive=False, retry=False)
+
+      if not response:
+        self.logger.warning(
```
illfelder: This seems like a non-exceptional case. Perhaps it should be a standard logger.info?

```
self.logger.info('Authentication token not found. Attempting unauthenticated download.')
```

google_compute_engine/metadata_watcher.py
```diff
@@ -153,14 +153,16 @@ def _GetMetadataUpdate(
...
+      self, metadata_key='', recursive=True, wait=True, timeout=None,
+      retry=True):
     """Wait for a successful metadata response.
 
     Args:
       metadata_key: string, the metadata key to watch for changes.
       recursive: bool, True if we should recursively watch for metadata changes.
       wait: bool, True if we should wait for a metadata change.
       timeout: int, timeout in seconds for returning metadata output.
+      retry: bool, True if it should keep retrying after failing.
```
illfelder: ```True if we should retry on failure.```

google_compute_engine/metadata_watcher.py
```diff
@@ -178,6 +180,10 @@ def _HandleMetadataUpdate(
           exception = e
           self.logger.error('GET request error retrieving metadata. %s.', e)
 
+      if not retry:
+        # If it didn't returned up to this point, and exception occured.
+        break
```
illfelder: I don't think the following is possible in the current code, but it's conceivable that `exception` will eventually be set, in which case the continue will be triggered before this break. To prevent that, I think you should move the if statement to the first line of the except (and you can drop the comment).

google_compute_engine/metadata_scripts/tests/script_retriever_test.py
```diff
@@ -33,47 +36,95 @@ def setUp(self):
...
+      self, mock_GetMetadata, mock_request, mock_tempfile):
+    auth_url = 'https://storage.googleapis.com/fake/url'
+    token_url = 'http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/token'
     mock_tempfile.return_value = mock_tempfile
     mock_tempfile.name = self.dest
-    mock_call.side_effect = [0, subprocess.CalledProcessError(1, 'Test')]
-    self.assertIsNone(self.retriever._DownloadGsUrl(gs_url, self.dest_dir))
+    self.retriever.token = None
+
+    mock_GetMetadata.return_value = {'token_type':'foo', 'access_token':'bar'}
```
illfelder: Please keep a space after the `:`.

google_compute_engine/metadata_scripts/tests/script_retriever_test.py
```diff
@@ -33,47 +36,95 @@ def setUp(self):
...
+
   @mock.patch('google_compute_engine.metadata_scripts.script_retriever.tempfile.NamedTemporaryFile')
-  def testDownloadGsUrlProcessError(self, mock_tempfile, mock_call):
-    gs_url = 'gs://fake/url'
+  @mock.patch('google_compute_engine.metadata_scripts.script_retriever.urlrequest.Request')
+  @mock.patch('google_compute_engine.metadata_watcher.MetadataWatcher.GetMetadata')
+  def testDownloadAuthUrlExceptionAndToken(
+      self, mock_GetMetadata, mock_request, mock_tempfile):
+    auth_url = 'https://storage.googleapis.com/fake/url'
+    token_url = 'http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/token'
```
illfelder: This is well over 80 chars.

illfelder: Note: only imports can be over 80 characters (or long URLs in comments).

gut:
Hi @illfelder , please review last commits. With those changes I now see metadata-scripts working.

illfelder:
I did leave review comments around 45 minutes ago. Are they not showing up for you or perhaps you replied to them in a new commit that hasn't been pushed?

google_compute_engine/metadata_watcher.py
gut: I see the advantages of having it as I wrote as the (desirable) error message `GET request ...` will be displayed on failure.

I don't think the `continue` can be reached, as when `retry == False`, the `exception` variable will be considered dead code. I explain: The `continue` on this function will never be triggered as no exception class defined on `except` clause is of type `NoneType`. Loop will reach a break after that, leaving no possibility, in this case, for continuing the `while` loop.

Do you agree in keeping it like I proposed or do you want me to change?

illfelder: I believe that you are correct that in the current code the continue cannot be hit, but I also think it is possible that behavior may change in the future. I do agree that the logging is desirable - great catch. How about we update the logic to this:

```
if not isinstance(e, type(exception)):
  exception = e
  self.logger.error('GET request error retrieving metadata. %s.', e)
if retry:
  continue
else:
  break
```

gut:
> I did leave review comments around 45 minutes ago. 

Sorry, I just noticed my comment was outdated now. I had it written just waiting for the tests to pass in order to send them. In the meantime you've sent the comments and I didn't re-read my text.

gut:
ok, looking good for me now. I hope you don't mind the small nits I also pushed with last comment fixes.

google_compute_engine/metadata_scripts/script_retriever.py
```diff
@@ -50,29 +57,38 @@ def _DownloadGsUrl(self, url, dest_dir):
...
-    self.logger.info('Downloading url from %s to %s using gsutil.', url, dest)
+    self.logger.info(
+        'Downloading url from %s to %s using authentication token.', url, dest)
+
+    if not self.token:
+      response = self.watcher.GetMetadata(
+          self.token_metadata_key, recursive=False, retry=False)
+
+      if not response:
+        self.logger.info('%s %s', 'Authentication token not found.',
```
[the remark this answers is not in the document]
gut: I'll implement it as you pointed out, still I'm concerned for the danger pointed out at the end of this answer: https://stackoverflow.com/a/18842890/5013070

gut:
great, thanks! Looking forward to see the new metadata-scripts working.

PR #648, opened 2018-08-29
Travis CI: Add Python 3.7 to the testing

cclauss:
* Add Python 3.7 to the testing In alignment with travis-ci/travis-ci#9069
* Lint on Python 3 instead of Python 2 because it is stricter.
* Fixed __os.errno__ --> __errno__ as discussed on the last bullet of:
    * https://docs.python.org/3/whatsnew/3.7.html#changes-in-the-python-api

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,10 @@ python:
 - 3.5
 - 3.6
 - pypy
+matrix:
+  include:
+    - python: 3.7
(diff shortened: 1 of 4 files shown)


cclauss:
Should we add a commitment of support for Python 3.7 at:
* https://github.com/GoogleCloudPlatform/compute-image-packages/blob/master/setup.py#L66

Or is that a separate PR?

illfelder:
Feel free to add a statement of support as part of this PR.

PR #655, opened 2018-09-07
oslogin: build libnss with debugging symbols

lucaskanashiro:
The GCC -g flag was added to generate debug information to be
used by GDB debugger.


lucaskanashiro:
I signed it!

lucaskanashiro:
I signed it!

PR #668, opened 2018-10-12
Add helper for setting per interface accept_ra_rt_info_max_plen

creatiwit:
accept_ra_rt_info_max_plen is set on a per device basis as per
https://github.com/torvalds/linux/blob/master/net/ipv6/ndisc.c#L1441


creatiwit:
Closing to add proper cla

PR #672, opened 2018-10-15
Add prefix 'google' to scripts inside 'scripts' dir

lucaskanashiro:
The scripts inside 'scripts' directory have general names that make
difficult identify from which package they belong to.


lucaskanashiro:
I signed it!

PR #674, opened 2018-10-19
Add journald config to forward logs to console

lucaskanashiro:
This is an alternative to rsyslog for systems that runs systemd.

I executed the image tests with this new config instead of rsyslog one and it worked.


xueweiz:
Seems the ForwardToConsole option would have performance impact on systemd-journald. This is still an open issue at:
https://github.com/systemd/systemd/issues/2815

PR #682, opened 2018-11-07
google-network-daemon: Start before cloud-init runs

Conan-Kudo:
The systemd `PartOf=` declaration here isn't enough to make cloud-init wait for the service to finish configuration, so we ensure that this runs before cloud-init is started to configure the network properly.

Signed-off-by: Neal Gompa <ngompa13@gmail.com>

### DIFF
--- a/google_compute_engine_init/systemd/google-network-daemon.service
+++ b/google_compute_engine_init/systemd/google-network-daemon.service
@@ -1,5 +1,6 @@
 [Unit]
 Description=Google Compute Engine Network Daemon
+Before=cloud-init.service
 After=local-fs.target network-online.target network.target rsyslog.service
 After=google-instance-setup.service
 Wants=local-fs.target network-online.target network.target


Conan-Kudo:
This is currently pending some testing. I'll remove the WIP marker when testing is complete.

Conan-Kudo:
Confirmed that it works as intended.

Conan-Kudo:
@illfelder Could you please take a look at this? It fixes a critical problem with trying to use these with creating Fedora images.

illfelder:
Hi, sorry for the delay. The Linux guest environment is not intended to be run (as is) with cloud init. There is functionality overlap and we don't want to create the perception that this use case is fully supported upstream. Please take a look at the Canonical provided guest environment (which repackages this code and runs alongside cloud init).

illfelder:
Canonical ships their own version of the `compute-image-packages` guest environment separate from cloud init:
https://packages.ubuntu.com/xenial-updates/gce-compute-image-packages

Conan-Kudo:
@illfelder Yes, we are in Fedora, too: https://src.fedoraproject.org/rpms/google-compute-engine-tools/blob/master/f/google-compute-engine-tools.spec

The problem is that it doesn't actually run in the correct order when both services are enabled, and it causes cloud-init to not be functional.

illfelder:
The code in the ubuntu repository is a fork that includes updates to improve behaviors when run in Ubuntu environments (with cloud-init). They include patches/disable functionality as appropriate. When cloud-init is running, portions of the guest should be disabled (for instance, if I recall correctly, host key generation is already done by cloud-init). If you are running an environment with cloud-init, startup scripts are also generally executed by cloud-init automatically so you should be able to disable the startup scripts functionality of this guest environment.

Conan-Kudo:
The only patching they do is to set up ordering for `snapd` and `cloud-final.service`: https://git.launchpad.net/~ubuntu-core-dev/+git/gce-compute-image-packages/tree/debian/patches?h=ubuntu/bionic

Otherwise, I have no idea what you're talking about...

PR #683, opened 2018-11-08
Fix SELinux module for OSLogin

hopkiw:
Move from the binary-format to the CIL format for SELinux module; this makes it a version-agnostic installation.


hopkiw:
tested build and install on 7.4 SAP image with SELinux enforcing; confirmed persistent through reboot.

PR #684, opened 2018-11-09
Two-factor updates: OS Login control file and metadata watcher

hopkiw:
Updates for two-factor auth in OSLogin

Some notes on the control file refactoring:
Rather than a litany of '|| exit 1' snippets, I've selectively enabled exit-on-error (set -e) in appropriate functions. Any failure to apply the changes should immediately be rolled back so as not to leave hosts in a broken state.

Pam and sshd configs are complex, so we take a full backup of these rather than trying to restore through modifications. 

For pam modification, this currently supports Debian and COS; other distros to come. Using here a strategy of appending our option to the end of the stack, except for authentication, where we get the first chance. The module as now never returns anything but success or failure; failure is ignored and the rest of the common-auth stack would be attempted. We could make this a hard failure by changing to `[success=done ignore=ignore default=bad]`. 

For sshd modification, we modify the option in place if it exists but is the wrong value, otherwise we append at bottom of file. 

For service restarts, we were treating the failure of a present control script (systemctl/service/invoke-rc.d) as something that means "try another method", when really it means "service isn't installed or enabled". We now treat that as a passing condition, whereas before we may have ended up starting a service when it was previously stopped. Now we check for the existence of three methods and if the method is available, we don't fall through to others.

Most of the freeBSD control paths weren't necessary after the other changes.

### DIFF
--- a/google_compute_engine/accounts/oslogin_utils.py
+++ b/google_compute_engine/accounts/oslogin_utils.py
@@ -38,18 +38,17 @@ class OsLoginUtils(object):
     self.oslogin_installed = True
     self.update_time = 0
 
-  def _RunOsLoginControl(self, action):
+  def _RunOsLoginControl(self, params):
     """Run the OS Login control script.
 
     Args:
-      action: str, the action to pass to the script
-          (activate, deactivate, or status).
+      params: list, the params to pass to the script
 
     Returns:
       int, the return code from the call, or None if the script is not found.
     """
     try:
-      return subprocess.call([constants.OSLOGIN_CONTROL_SCRIPT, action])
+      return subprocess.call([constants.OSLOGIN_CONTROL_SCRIPT] + params)
     except OSError as e:
       if e.errno == errno.ENOENT:
         return None
@@ -62,7 +61,7 @@ class OsLoginUtils(object):
     Returns:
       bool, True if OS Login is installed.
     """
-    retcode = self._RunOsLoginControl('status')
+    retcode = self._RunOsLoginControl(['status'])
     if retcode is None:
       if self.oslogin_installed:
         self.logger.warning('OS Login not installed.')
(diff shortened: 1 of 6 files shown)


google_compute_engine_oslogin/bin/google_oslogin_control
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
illfelder: This might cause problems on bsd which doesn't have bash by default.

google_compute_engine_oslogin/bin/google_oslogin_control
```diff
@@ -13,254 +13,300 @@
...
+fi
 
-get_status() {
-  if grep "^passwd:" ${nss_config} | grep -q "oslogin"; then
-    exit 0
-  else
-    exit 3
-  fi
-}
+# Main
```
illfelder: These comments make me sad.

google_compute_engine/accounts/oslogin_utils.py
```diff
@@ -123,7 +123,7 @@ def UpdateOsLogin(self, enable, duration=NSS_CACHE_DURATION_SEC):
     self.update_time = current_time
     if enable:
       self.logger.info('Activating OS Login.')
-      return self._RunOsLoginControl('activate') or self._RunOsLoginNssCache()
+      return self._RunOsLoginControl('activate --twofactor' if two_factor else 'activate') or self._RunOsLoginNssCache()
     else:
       self.logger.info('Deactivating OS Login.')
       return (self._RunOsLoginControl('deactivate')
```
illfelder: We should update this function to take a string and a boolean for 2fa. It would be a lot more readable and easier to maintain in the future if we add other options.

google_compute_engine_oslogin/bin/google_oslogin_control
hopkiw: i'll need to do some BSD testing then because some of the code here is probably bash-specific.

hopkiw: i have begrudgingly restored sh(1) compatible syntax

PR #698, opened 2018-11-30
Update two factor enablement on change.

illfelder:


google_compute_engine/accounts/oslogin_utils.py
```diff
@@ -107,17 +113,24 @@ def UpdateOsLogin(self, enable, two_factor=False):
     Returns:
       int, the return code from updating OS Login, or None if not present.
     """
-    status = self._GetStatus()
-    if status is None:
+    oslogin_status = self._GetStatus(two_factor=False)
+    two_factor_status = self._GetStatus(two_factor=True)
```
hopkiw: you don't need to do two checks, following #700 you can simple add two_factor=True if needed and get a true/false response

illfelder: If OS Login Two Factor is set to false in metadata, we still need to know whether OS Login and OS Login two factor were enabled previously. There's multiple similar cases. We end up needing to know both states.

illfelder: Updated the logic. Thanks for the feedback.

google_compute_engine/accounts/oslogin_utils.py
```diff
@@ -105,28 +111,29 @@ def UpdateOsLogin(self, enable, two_factor=False):
...
-    if enable:
-      self.logger.info('Activating OS Login.')
-      params = ['activate']
-      if two_factor:
-        params += ['--twofactor']
-      return self._RunOsLoginControl(params) or self._RunOsLoginNssCache()
-    else:
+    elif status:
       self.logger.info('Deactivating OS Login.')
       return (self._RunOsLoginControl(['deactivate'])
```
hopkiw: not clear why this is an or.. even tho i think it's my fault -g-

google_compute_engine/accounts/oslogin_utils.py
```diff
@@ -105,28 +111,29 @@ def UpdateOsLogin(self, enable, two_factor=False):
...
+    # Two factor can only be enabled when OS Login is enabled.
+    two_factor = two_factor and enable
+    status = self._GetStatus(two_factor=two_factor)
     if status is None:
       return None
 
-    current_time = time.time()
-    if status == enable:
-      if status and current_time - self.update_time > NSS_CACHE_DURATION_SEC:
+    if enable:
```
hopkiw: can we change the variable names to something more descriptive?

illfelder: Sure - made it `enable_oslogin`.

PR #699, opened 2018-12-03
Accounts: separate gpasswd command in two

lucaskanashiro:
Specify different commands to add and remove users from a group.
In FreeBSD, the flags used by pw utility are not the same used
by gpasswd.


illfelder:
This does not update the template config file generation or the readme.

lucaskanashiro:
Sorry, I forgot that. It is updated with the required modifications.

PR #702, opened 2018-12-04
Fix regression on google_oslogin_control FreeBSD compatibility

lucaskanashiro:


google_compute_engine_oslogin/bin/google_oslogin_control
```diff
@@ -184,6 +184,16 @@ modify_pam_sshd() (
     session_end=$($sed -n '/^session/=' "$pam_config" | tail -1)
     $sed -i"" "${session_end}a ${added_config}" "$pam_config"
   fi
+
+  # In FreeBSD, the used flags are not supported, replacing them with the
+  # previous ones (requisite and optional).
+  if is_freebsd; then
+    ${sed} -ri 's#(account)\s+\[.+\]\s+(pam_oslogin_login.so)#\1\t requisite\t\2#' "${pam_config}"
```
hopkiw: instead of edit and then edit-again, it may  be better to use an if-freebsd changing variables similar to how el6 is handled in the modify_sshd() function..

lucaskanashiro: That makes sense, already fixed.

PR #703, opened 2018-12-04
Restart systemd-logind

hopkiw:
systemd-logind seems to cache the presence or absence of oslogin_cache; this causes a chain of failure from logind->pam_systemd->pam(sshd:session) that gives false negatives in logs and prevents a logind session including XDG_SESSION_ID/XDG_RUNTIME_DIR from being set up. fix is simply to restart systemd-logind if present.

additionally, update policy to permit google-users.d creation and changes


google_compute_engine_oslogin/bin/google_oslogin_control
```diff
@@ -243,9 +243,9 @@ restart_sshd() {
   done
 }
 
-restart_nscd() {
-  echo "Restarting NSCD, if present"
-  for svc in "nscd" "unscd"; do
+restart_svcs() {
+  echo "Restarting optional services.."
```
illfelder: Extra period.

hopkiw: fixxed

PR #707, opened 2018-12-05
Fix logic for enabling OS Login two factor.

illfelder:


google_compute_engine/accounts/oslogin_utils.py
```diff
@@ -103,37 +103,50 @@ def _RemoveOsLoginNssCache(self):
         if e.errno != errno.ENOENT:
           raise
 
-  def UpdateOsLogin(self, enable_oslogin, two_factor=False):
+  def UpdateOsLogin(self, oslogin_desired, two_factor_desired=False):
     """Update whether OS Login is enabled and update NSS cache if necessary.
 
     Args:
-      enable_oslogin: bool, enable OS Login if True, disable if False.
+      oslogin_desired: bool, enable OS Login if True, disable if False.
```
hopkiw: two_factor_desired is an undocumented arg

google_compute_engine/accounts/oslogin_utils.py
```diff
@@ -103,37 +103,50 @@ def _RemoveOsLoginNssCache(self):
...
+      # OS Login is desired and not enabled.
+      if not oslogin_configured:
         self.logger.info('Activating OS Login.')
-        params = ['activate']
-        if two_factor:
-          params += ['--twofactor']
         return self._RunOsLoginControl(params) or self._RunOsLoginNssCache()
-      # OS Login is already enabled. Update the cache if appropriate.
+      # Enable two factor authentication.
+      if not two_factor_configured and two_factor_desired:
```
hopkiw: rather than aiming for symmetry between this and the below, maybe optimize for human language parsing e.g. `if two_factor_desired and not two_factor_configured:`

google_compute_engine/accounts/oslogin_utils.py
```diff
@@ -103,37 +103,50 @@ def _RemoveOsLoginNssCache(self):
...
+        self.logger.info('Activating OS Login with two factor enabled.')
+        return self._RunOsLoginControl(params) or self._RunOsLoginNssCache()
```
hopkiw: this `or` is deeply confusing and only works because we pass the unix return code from `_RunOsLoginControl()` which parses as false. explicitly running these functions would be much easier to parse

google_compute_engine/accounts/oslogin_utils.py
```diff
@@ -103,37 +103,50 @@ def _RemoveOsLoginNssCache(self):
...
+      if not oslogin_configured:
         self.logger.info('Activating OS Login.')
-        params = ['activate']
-        if two_factor:
-          params += ['--twofactor']
         return self._RunOsLoginControl(params) or self._RunOsLoginNssCache()
-      # OS Login is already enabled. Update the cache if appropriate.
+      # Enable two factor authentication.
+      if not two_factor_configured and two_factor_desired:
+        self.logger.info('Activating OS Login with two factor enabled.')
```
hopkiw: this log string is different from above, but the above may also have two factor enabled.

google_compute_engine/accounts/oslogin_utils.py
illfelder: Thanks!

google_compute_engine/accounts/oslogin_utils.py
illfelder: Done.

google_compute_engine/accounts/oslogin_utils.py
illfelder: Made it more clear. Thanks!

google_compute_engine/accounts/oslogin_utils.py
illfelder: Short circuiting when a command fails is preferred so I'll leave it as is.

PR #708, opened 2018-12-05
Add unit test coverage for each enablement state.

illfelder:


google_compute_engine/accounts/tests/oslogin_utils_test.py
```diff
@@ -318,36 +249,79 @@ def testUpdateOsLoginUpdateCache(self, mock_time):
...
+      mocks.attach_mock(self.mock_logger, 'logger')
+      mocks.attach_mock(self.mock_oslogin, 'oslogin')
+      self.mock_oslogin._RunOsLoginControl.return_value = 0
+      self.mock_oslogin._GetStatus.side_effect = [
+          oslogin_config, two_factor_config]
+      mock_time.return_value = 6 * 60 * 60
+      return_value = oslogin_utils.OsLoginUtils.UpdateOsLogin(
+          self.mock_oslogin, oslogin, two_factor_desired=two_factor)
+
+      if oslogin_config:
```
hopkiw: i dont think i would be able to maintain this. would it be a lot of work to use a more traditional testing structure with explicitly called tests?

PR #719, opened 2018-12-20
WIP No c++11 without pam fix

rbalint:


rbalint:
I signed it!

packages/google-compute-engine-oslogin/utils/oslogin_utils.cc
```diff
@@ -23,10 +23,7 @@
 #include <iostream>
 #include <sstream>
 
-#if defined(__clang__) || __GNUC__ > 4 || \
-  (__GNUC__ == 4 && (__GNUC_MINOR__ > 9 || \
-                     (__GNUC_MINOR__ == 9 && \
-                      __GNUC_PATCHLEVEL__ > 0)))
+#if __cplusplus >= 201103L
```
hopkiw: can you restore the if-__clang__ ? the cplusplus check seems only to satisfy the GNUC version check

rbalint: @hopkiw I reworked the patch but indeed, cland may need to be treated differently.

PR #756, opened 2019-04-11
Fix platform detection in case of FreeBSD in compat module.

dwich:
Problem: For platform detection for Python >= 3.6 the `distro` package is used. For lower version of Python the built-in package `platform` is used.

`distro.linux_distribution()` is used to detect Linux
`distro.version()` is used to detecting FreeBSD

Method `distro.version()` should not be used because `distro` can be either `distro` itself or `import platform as distro`, therefore:

`platform.version()` and `distro.version()` return different values.
`platform.version()` returns 'FreeBSD 11.2-RELEASE-p9.....'
`distro.version()` returns '11.2'

Solution: Do not detect platform and version (in case of FreeBSD) using `distro` package, which is useful for linux but it's not necessary for FreeBSD.

### DIFF
--- a/packages/python-google-compute-engine/google_compute_engine/compat.py
+++ b/packages/python-google-compute-engine/google_compute_engine/compat.py
@@ -25,11 +25,17 @@ else:
   import platform as distro
 
 if 'freebsd' in sys.platform:
-  distribution = distro.version().split()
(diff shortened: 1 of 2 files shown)


illfelder:
Thanks for the contribution! This looks good to me.

PR #761, opened 2019-04-25
Add support for writing host keys to guest attributes.

wrigri:


wrigri:
I don't think this is quite ready to merge yet. I would like to improve the unittests a bit.

PR #762, opened 2019-04-29
Add check for if google_sudoers file is empty

cjd:
Additional test added to existing test for missing file so can handle additional potential failure mode


hopkiw:
this codebase is deprecated, but this behavior is fixed in the new guest agent

PR #763, opened 2019-04-29
Add support for writing host keys to guest attributes.

wrigri:


packages/python-google-compute-engine/google_compute_engine/instance_setup/instance_setup.py
```diff
@@ -140,6 +153,27 @@ def _GenerateSshKey(self, key_type, key_dest):
...
+
+  def _WriteHostKeyToGuestAttributes(self, key_type, key_value):
+    """Write a host key to guest attributes, ignoring errors."""
+    headers = {'Metadata-Flavor': 'Google'}
+    url = GUEST_ATTRIBUTES_URL + key_type
+    req = PutRequest(url, key_value, headers)
+    try:
+      response = urlrequest.urlopen(req)
+      self.logger.debug(response)
+      self.logger.info('Wrote %s host key to guest attributes.' % key_type)
```
illfelder: For `logger.info` you should use a comma for variable substitution arguments.

packages/python-google-compute-engine/google_compute_engine/instance_setup/instance_setup.py
```diff
@@ -28,9 +28,20 @@
...
 from google_compute_engine.instance_setup import instance_config
 
 
+class PutRequest(urlrequest.Request):
+    def get_method(self):
+        return 'PUT'
+
+
+GUEST_ATTRIBUTES_URL = ('http://metadata.google.internal/computeMetadata/v1beta1/'
+                        'instance/guest-attributes/hostkeys/')
```
illfelder: Let's separate out the `hostkeys/` suffix.

packages/python-google-compute-engine/google_compute_engine/instance_setup/instance_setup.py
```diff
@@ -122,6 +133,8 @@ def _GenerateSshKey(self, key_type, key_dest):
     Args:
       key_type: string, the type of the SSH key.
       key_dest: string, a file location to store the SSH key.
+    Returns:
```
illfelder: Please add a new line between the args and return sections.

packages/python-google-compute-engine/google_compute_engine/instance_setup/instance_setup.py
```diff
@@ -140,6 +153,27 @@ def _GenerateSshKey(self, key_type, key_dest):
...
+    with open('%s.pub' % key_dest, 'r') as pk:
+        key_data = pk.read()
+
+    key_values = key_data.split()
+    if len(key_values) < 2:
+        return
+    else:
+        return key_values[0], key_values[1]
+
+  def _WriteHostKeyToGuestAttributes(self, key_type, key_value):
```
illfelder: We can make a new module for writing metadata since this is generally usable. It's fine to leave it as-is until the feature is GA.

wrigri: Yes, as we discussed, it's a good idea to split this out, but probably better to wait until GA. Once guest attributes is GA we can create a module for writing those.

packages/python-google-compute-engine/google_compute_engine/instance_setup/instance_setup.py
wrigri: Done.

packages/python-google-compute-engine/google_compute_engine/instance_setup/instance_setup.py
wrigri: Done

packages/python-google-compute-engine/google_compute_engine/instance_setup/instance_setup.py
wrigri: Done.

packages/python-google-compute-engine/google_compute_engine/instance_setup/instance_setup.py
```diff
@@ -28,9 +28,21 @@
...
 
 
+class PutRequest(urlrequest.Request):
+    def get_method(self):
+        return 'PUT'
+
+
+GUEST_ATTRIBUTES_URL = ('http://metadata.google.internal/computeMetadata/v1beta1/'
+                        'instance/guest-attributes')
+GA_NAMESPACE = 'hostkeys'
```
illfelder: `HOSTKEY_NAMESPACE`

packages/python-google-compute-engine/google_compute_engine/instance_setup/instance_setup.py
```diff
@@ -140,6 +155,27 @@ def _GenerateSshKey(self, key_type, key_dest):
...
     file_utils.SetPermissions(key_dest, mode=0o600)
     file_utils.SetPermissions('%s.pub' % key_dest, mode=0o644)
+    with open('%s.pub' % key_dest, 'r') as pk:
+        key_data = pk.read()
+
+    key_values = key_data.split()
+    if len(key_values) < 2:
+        return
+    else:
+        return key_values[0], key_values[1]
```
hopkiw: key_type is passed in; why return it?

packages/python-google-compute-engine/google_compute_engine/instance_setup/instance_setup.py
```diff
@@ -140,6 +155,27 @@ def _GenerateSshKey(self, key_type, key_dest):
 
     file_utils.SetPermissions(key_dest, mode=0o600)
     file_utils.SetPermissions('%s.pub' % key_dest, mode=0o644)
+    with open('%s.pub' % key_dest, 'r') as pk:
+        key_data = pk.read()
+
+    key_values = key_data.split()
+    if len(key_values) < 2:
+        return
```
hopkiw: when is this the case and is it OK to just ignore it..?

packages/python-google-compute-engine/google_compute_engine/instance_setup/instance_setup.py
```diff
@@ -28,9 +28,21 @@
 from google_compute_engine import logger
 from google_compute_engine import metadata_watcher
 from google_compute_engine.boto import boto_config
+from google_compute_engine.compat import urlerror
+from google_compute_engine.compat import urlrequest
 from google_compute_engine.instance_setup import instance_config
 
 
+class PutRequest(urlrequest.Request):
+    def get_method(self):
```
hopkiw: this doesn't seem necessary.. `get_method()` will return 'PUT' if there's data, and we always pass data in to this object.

illfelder: The new indentation is wrong. The code should conform to Google Python style with 2 indents (4 for line continuations).

wrigri: No, if you pass in data, get_method() returns a 'POST' and guest attributes needs a 'PUT'.

packages/python-google-compute-engine/google_compute_engine/instance_setup/instance_setup.py
wrigri: In theory it shouldn't ever be the case, but if there is something weird in the file, or if it's empty or something, I just return None and ignore it. As for whether it's ok to ignore stuff, for the near future this is designed to fail in a way that doesn't affect current behavior, so if things don't work as expected we just want to continue.

packages/python-google-compute-engine/google_compute_engine/instance_setup/instance_setup.py
wrigri: Sanity check. The passed in key_type is used to find the right file to read, while the value that is returned is from the contents of the file. This shouldn't ever be different, and I don't have a strong attachment to doing it this way, so if you would prefer that this only return the key value, I'm ok with that.

packages/python-google-compute-engine/google_compute_engine/instance_setup/instance_setup.py
wrigri: >>> data = "asdfasdf"
>>> r = urllib2.Request("http://foo.com", data)
>>> r.get_method()
'POST'

wrigri: Fixed. (I did some development on VMs and vi on those VMs defaults to 4 space indents. Some of it came over when I copied from the VMs.)

packages/python-google-compute-engine/google_compute_engine/instance_setup/instance_setup.py
wrigri: Done.

packages/python-google-compute-engine/google_compute_engine/instance_setup/instance_setup.py
```diff
@@ -180,7 +216,9 @@ def _SetSshHostKeys(self, host_key_types=None):
       for key_file in set(key_files) | set(key_types_files):
         key_type = file_regex.match(key_file).group('type')
         key_dest = os.path.join(key_dir, key_file)
-        self._GenerateSshKey(key_type, key_dest)
+        key_data = self._GenerateSshKey(key_type, key_dest)
+        if key_data:
+            self._WriteHostKeyToGuestAttributes(key_data[0], key_data[1])
```
illfelder: Two spaces.

packages/python-google-compute-engine/google_compute_engine/instance_setup/instance_setup.py
hopkiw: i'm not opposed, just unclear why you'd return something you're passing in. maybe add a comment explaining but otherwise your explanation is good enough for me.

packages/python-google-compute-engine/google_compute_engine/instance_setup/instance_setup.py
wrigri: Done.

PR #765, opened 2019-05-13
Fix XPS settings with more than 64 vCPUs.

braffert:


zmarano:
Ehh... can you rebase against development --- my attempt to fix that didn't yield amazing results :)

PR #771, opened 2019-05-20
Retry download for metadata scripts

dntczdx:
script_retriever is on the critical path of instance boot. To improve the success rate, let's retry downloading when failure happens.

### DIFF
--- a/packages/python-google-compute-engine/google_compute_engine/metadata_scripts/script_retriever.py
+++ b/packages/python-google-compute-engine/google_compute_engine/metadata_scripts/script_retriever.py
@@ -15,17 +15,48 @@
 
 """Retrieve and store user provided metadata scripts."""
 
-import ast
+import functools
 import re
 import socket
-import subprocess
 import tempfile
+import time
 
 from google_compute_engine import metadata_watcher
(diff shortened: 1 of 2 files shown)


packages/python-google-compute-engine/google_compute_engine/metadata_scripts/script_retriever.py
```diff
@@ -192,8 +223,10 @@ def _GetAttributeScripts(self, attribute_data, dest_dir):
     metadata_value = attribute_data.get(metadata_key)
     if metadata_value:
       self.logger.info('Found %s in metadata.', metadata_key)
-      script_dict[metadata_key] = self._DownloadScript(
-          metadata_value, dest_dir)
+      downloaded_dest = self._DownloadScript(metadata_value, dest_dir)
+      if downloaded_dest is None:
+        self.logger.warning('Failed to download metadata script.')
```
illfelder: Tests don't hit this case.

dntczdx: Indeed! I missed this...

dntczdx:
Added a UT to cover new warning for download failure

PR #794, opened 2019-06-11
New package versioning

zmarano:
- Move to YYYYMMDD.0-r versioning.
- Open to suggestions, ideas, and other variations.

### DIFF
--- a/daisy_workflows/build_deb_packages.sh
+++ b/daisy_workflows/build_deb_packages.sh
@@ -56,7 +56,13 @@ done
 
 # Copy the deb and dsc files to the output.
 cd "${workdir}/output"
-gsutil cp *.dsc *.deb ${OUTPUT}
+# For Debian 10 right now, only copy OS Login package.
+if [[ "$(cut -d. -f1 </etc/debian_version)" == "10" ]]; then
+  gsutil cp google-compute-engine-oslogin*.* ${OUTPUT}
+else
+  gsutil cp *.dsc *.deb ${OUTPUT}
+fi
+
 if [ $? -ne 0 ]; then
   echo "BuildFailed: copying to ${OUTPUT} failed."
   exit 1
--- a/daisy_workflows/build_debian.wf.json
+++ b/daisy_workflows/build_debian.wf.json
@@ -20,8 +20,14 @@
     "setup-disk": {
       "CreateDisks": [
         {
-          "Name": "disk-deb-build",
+          "Name": "disk-deb9-build",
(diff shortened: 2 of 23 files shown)


hopkiw:
we need the following information associated with each built package:
when it was built
whether it is the google-built package
what code it was built from (e.g. git commit)

a version moreover must serve the purpose of allowing newer software to replace older software, and must be parseable by RPM, Googet, and DPKG systems.

The date idea makes sense, especially combined with a (two-digit) build number for that day. the git commit does not need to be in the version if there is another place for it in the package system.

for the proposal, i don't understand what the 'r' refers to. otherwise i think the build number should be made two digit and we should be clear the 'r' is in the release, not in the version. we can use release to differentiate between google and distro/vendor if that's easier.

work that needs to be done: the proposal needs to be tested in all supported package managers

hopkiw:
testing: we have high confidence that 2019xxxxx is going to beat out any legacy semver style versions. we still want to test upgrades in our proposed scheme, and ensure google-authored packages always beat distro partner packages, if both are available.

hopkiw:
some distro partners are also using this date format for major version, so comparisons can be strange. if the date for distro package is just the release date, their package with older code may appear 'newer'. even worse, we may end up with leapfrogging code where the actual code becomes older, then newer, then older, so on. 
there may be some solutions to this.. debian provides `epoch`, which has higher priority than version. there are some lexically-higher characters like ~ we could prepend.

hopkiw:
epoch looks promising. leading characters and tildes in the version also seem to work, at least for debian. keeping dpkg, googet and rpm formats as similar as possible is worthwhile.

hopkiw:
per discussion, we think we have a good 'get the latest guest environment' story for customers. we will produce packages for all distros that are always newer than what the distro provides, so users can simply add repo + upgrade packages. this means new SuSE and Ubuntu packages to build on our end. Bonus is that we can run these through tests as well and get out ahead of release issues.

hopkiw:
zmarano confirms testing is done, epoch does what we want. git info can go in the packages as follows.

RPM packages:
`Vcs: <value>` in RPM spec file
retrieved with `rpm -q --qf '%{VCS}\n' <packagename>`

Debian packages:
 `X-BCS-VCS: <value>` in the `debian/control` file
retrieved with `dpkg-query -W -f '${VCS}\n' <packagename>`

Value is TBD but can be git short or long commit hash, github URL for the commit, github URL for the tree at that commit, or github URL for a 'release' tarball.

PR #796, opened 2019-06-13
google_oslogin_control: make it FreeBSD compatible

lucaskanashiro:
These modifications make OSLogin work again on FreeBSD images. They also fix some warnings.

### DIFF
--- a/packages/google-compute-engine-oslogin/google_oslogin_control
+++ b/packages/google-compute-engine-oslogin/google_oslogin_control
@@ -201,6 +201,9 @@ modify_pam_config() (
     # Get location of system-remote-login.
     insert=$($sed -rn "/^auth\s+include\s+system-remote-login/=" "$pam_sshd_config")
     # TODO: find su_insert point for arch linux.
+  elif is_freebsd; then
+    # Get location of the first auth occurrence
+    insert=$($sed -rn '/^auth/=' "$pam_sshd_config" | head -1)
   fi
 
   added_config="$added_comment"
@@ -223,7 +226,7 @@ modify_pam_config() (
   # Insert su blocker at top of `su:account` stack.
   if [ -n "$su_insert" ] && ! grep -qE "$pam_account_su" "$pam_su_config"; then
     added_su_config="${added_comment}\n${pam_account_su}"
-    sed -i"" "${su_insert}i ${added_su_config}" "$pam_su_config"
+    $sed -i"" "${su_insert}i ${added_su_config}" "$pam_su_config"
   fi
 
   # Append account modules at end of `sshd:account` stack.
@@ -255,6 +258,12 @@ restore_pam_config() {
 }
 
 modify_group_conf() {
+  # In FreeBSD there is no pam_group config file similar to
+  # /etc/security/group.conf.
+  if is_freebsd; then
+    return
+  fi
+
   local group_config="${1:-${group_config}}"
   local group_conf_entry="sshd;*;*;Al0000-2400;adm,dip,docker,lxd,plugdev,video"
 
@@ -264,6 +273,12 @@ modify_group_conf() {
 }
 
 restore_group_conf() {
+  # In FreeBSD there is no pam_group config file similar to
+  # /etc/security/group.conf.
+  if is_freebsd; then
+    return
+  fi
+
   local group_config="${1:-${group_config}}"
 
   $sed -i"" "/${added_comment}/{n;d}" "$group_config"


packages/google-compute-engine-oslogin/google_oslogin_control
```diff
@@ -347,18 +361,34 @@ remove_google_dirs() {
 }
 
 activate() {
-  for func in modify_sshd_conf modify_nsswitch_conf \
-              modify_pam_config setup_google_dirs restart_svcs restart_sshd \
-              modify_group_conf; do
+  if is_freebsd; then
```
hopkiw: the only difference here is one function.. i think that function can just short-circuit in freebsd case, instead.

packages/google-compute-engine-oslogin/google_oslogin_control
```diff
@@ -314,16 +320,24 @@ restart_sshd() {
     return 0
   fi
   echo "Restarting SSHD"
-  for svc in "ssh" "sshd"; do
-    restart_service "$svc"
-  done
+  if is_freebsd; then
```
hopkiw: 'ssh' doesnt exist most of the time, either, we just try it in case. does it have a negative effect in freebsd?

packages/google-compute-engine-oslogin/google_oslogin_control
```diff
@@ -154,6 +154,7 @@ modify_pam_config() (
     pam_account_oslogin="account    optional pam_oslogin_admin.so"
     pam_account_admin="account    requisite pam_oslogin_login.so"
     pam_session_homedir="session    optional pam_mkhomedir.so"
+    pam_account_su="account    optional pam_oslogin_login.so"
```
hopkiw: in linux we have mapped success to 'bad', so that oslogin users will be *prevented* from su. this line won't have the same effect

packages/google-compute-engine-oslogin/google_oslogin_control
lucaskanashiro: Yes, it does. If we try to restart the `ssh` service the script returns non zero, which will not allow OSLogin activation.

packages/google-compute-engine-oslogin/google_oslogin_control
lucaskanashiro: Right, this is a better solution.

packages/google-compute-engine-oslogin/google_oslogin_control
hopkiw: what script returns non zero? we try systemd, `service` and then init scripts. but in each case we check for existence and status first so that non-existent services should have no impact. if you can share the specific details i could probably make an improvement here

lucaskanashiro: Sorry for the delay, I forgot to reply this message. The problem here is that the `restart_service` function does not handle well FreeBSD services, explanation below:

```
if ! service "$service" status 2>&1 | grep -q unrecognized; then
    service "$service" restart
    return $?
else
    return 0
fi
```
The snippet above try to find the `unrecognized` string in the output but this string does not exist in FreeBSD output when trying to check the status of a non existent service:

```
kanashiro@kanashiro-freebsd-13-debug:~ $ sudo service ssh status
ssh does not exist in /etc/rc.d or the local startup
directories (/usr/local/etc/rc.d), or is not executable
kanashiro@kanashiro-freebsd-13-debug:~ $ echo $?
1
kanashiro@kanashiro-freebsd-13-debug:~ $ sudo service sshd status
sshd is running as pid 31038.
kanashiro@kanashiro-freebsd-13-debug:~ $ echo $?
0
```
The best approach would be to improve the 'restart_service` function indeed, any idea about the best way to do it? Check also for `does not exist` string?

lucaskanashiro: sorry, I was using the FreeBSD ports package which is synced with `master` and does not contain the modification I proposed above (it's already implemented in `development`). So, we need to wait for the next release to get it into FreeBSD package (now it is fixed via patch).

lucaskanashiro:
@hopkiw now I think this PR is ready to be merged.

packages/google-compute-engine-oslogin/google_oslogin_control
hopkiw: the simplest solution is to also check for 'does not exist', yeah. checking BSD [manpage][1] i see there are `-e` and `-r` arguments for explicit checking, we could check that list first. 

[1]: https://www.freebsd.org/cgi/man.cgi?query=service&manpath=FreeBSD+12.0-RELEASE+and+Ports

PR #799, opened 2019-07-03
Disk expand: Make sure the disk has settled before continuing.

zmarano:


packages/gce-disk-expand/src/expandfs-lib.sh
```diff
@@ -111,6 +111,8 @@ parted_needresize() {
     return 1
   fi
 
+  udevadm settle
```
hopkiw: this should be on line 113 of `src/expandfs-lib.sh`, right after the parted command. we might find out we need it in other places too, depending on how testing goes.

PR #808, opened 2019-07-11
Specify cpp version in Makefile

hopkiw:
Be explicit about cpp std in use.


packages/google-compute-engine-oslogin/src/Makefile
```diff
@@ -1,11 +1,12 @@
 SHELL = /bin/sh
 TOPDIR = $(realpath ..)
 
-VERSION = 20190708.00
+VERSION = $(shell date +%Y%m%d.00)
```
zmarano: Lets revert this back for now.

PR #823, opened 2019-07-31
Re-enable boto config and don't write plugin line.

### LINKED ISSUE #820: Re-enable boto setup on python3 but remove plugin line
For standalone gsutil installs (not the included gcloud installed gsutil) the boto config file is needed but without the problematic plugin which breaks on python3 systems. We can instead just write the rest of the boto config file and drop the plugin.

zmarano:
Fixed #820

### DIFF
--- a/packages/python-google-compute-engine/google_compute_engine/boto/boto_config.py
+++ b/packages/python-google-compute-engine/google_compute_engine/boto/boto_config.py
@@ -80,7 +80,6 @@ class BotoConfig(object):
     config = config_manager.ConfigManager(
         config_file=self.boto_config_template,
         config_header=self.boto_config_header)
-    boto_dir = os.path.dirname(self.boto_config_script)
 
     config.SetOption('GSUtil', 'default_project_id', project_id)
     config.SetOption('GSUtil', 'default_api_version', '2')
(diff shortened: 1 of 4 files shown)


packages/python-google-compute-engine/google_compute_engine/boto/boto_config.py
```diff
@@ -85,5 +86,7 @@ def _CreateConfig(self, project_id):
     config.SetOption('GSUtil', 'default_project_id', project_id)
     config.SetOption('GSUtil', 'default_api_version', '2')
     config.SetOption('GoogleCompute', 'service_account', 'default')
-    config.SetOption('Plugin', 'plugin_directory', boto_dir)
+    # Don't set the plugin for python3 as it is not compatible.
```
houglum: The version of Python you're using to run *this script* doesn't necessarily correspond to the version of Python that users will invoke to run code that relies on Boto (gsutil, for example, can be run with either Python 2 or 3).

As mentioned in our Hangouts discussion, I think we can just stop writing this plugin line entirely.  Gsutil doesn't need it. If we have any documentation that instructs people to use Boto, we should also update it to mention that users should install the google_compute_engine plugin and set up their boto config file (using the lines below) such that Boto loads the GCE oauth2 auth_handler.

zmarano: SGTM

PR #828, opened 2019-08-01
Use non-numeric project ID in generated boto.cfg for GSUtil

### LINKED ISSUE #827: Generated /etc/boto.cfg should use non-numeric project ID
The /etc/boto.cfg file that gets generated during instance configuration should use the project's ID, not its project number, for the `default_project_id` attribute under the `[GSUtil]` section of the config.  Most gsutil integration tests (i.e. requester-pays functionality, KMS key functionality, and other commands that require fetching the GCS_owned service account for the default project) fail when you use the project's number, as this value gets substituted into a URL where GCS expects the project ID, not the project number.  We get around this in our release tests by overriding the `default_project_id` attribute in the `$HOME/.boto` file, but we shouldn't require this manual step for our users.

As this attribute is under the `[GSUtil]` section of the config, no other programs should be relying on it, and we've tested that it's safe to change this by using our test projects' IDs instead of their numbers during gsutil prerelease testing.  If you know of any edge cases where other programs are relying on this value, please list them in this issue.  Otherwise, I'll send a PR shortly to make the generated config contain the project ID.

houglum:
...for the "default_project_id" attribute under the [GSUtil] section of
the config file at /etc/boto.cfg.

Fixes #827.


houglum:
@zmarano @illfelder Do you see any problems with switching `default_project_id` from the numeric value to the actual ID?

zmarano:
I don't see any real issues except that its been this way forever so why change it. But, if you think that it is the better method for gsutil, then its probably ok.

houglum:
The only thing using the number accomplishes is preventing us from using some gsutil functionality that expects the actual project ID.  When running our prerelease tests, we overwrite this value to contain the ID instead of the number, so we've been testing using this method for several years already.

packages/python-google-compute-engine/google_compute_engine/boto/boto_config.py
```diff
@@ -58,22 +58,23 @@ def __init__(self, project_id=None, debug=False):
...
+    """Get the (non-numeric) project ID for this VM.
 
     Returns:
-      string, the numeric project ID if one is found.
+      string, the project ID if one is found.
     """
-    project_id = 'project/numeric-project-id'
-    return self.watcher.GetMetadata(metadata_key=project_id, recursive=False)
+    project_id_key = 'project/project-id'
+    return self.watcher.GetMetadata(metadata_key=project_id_key,
```
illfelder: Please line wrap from the first open paren.

houglum: Done.

PR #833, opened 2019-08-07
Move oslogin to new repo

hopkiw:
Move the OS Login components to new repo `guest-oslogin`, leaving a README.


zmarano:
You'll probably want to fix the current package build workflows so they don't fail when they fail to find the expected build files.

PR #834, opened 2019-08-07
Detect hostname changes

hopkiw:
The 'right' way to detect hostname changes is to refresh the DHCP client. This varies by distro, so add a new distro-specific function. Also add a common systemctl command to helpers to reduce code duplication.


packages/python-google-compute-engine/google_compute_engine/networking/network_daemon.py
```diff
@@ -98,10 +103,13 @@ def HandleNetworkInterfaces(self, result):
...
-    for interface in network_interfaces:
-      if self.ip_forwarding_enabled:
-        self.ip_forwarding.HandleForwardedIps(
-            interface.name, interface.forwarded_ips, interface.ip)
+    if self.ip_forwarding_enabled:
+      for interface in network_interfaces:
+        self.ip_forwarding.HandleForwardedIps(interface.name,
+                                              interface.forwarded_ips,
+                                              interface.ip)
+    if socket.gethostname() != result['hostname']:
```
zmarano: Do we know for certain that we will always match the FQDN returned and now get into a loop where it will not match ever? I could imagine this returning values that may never match depending on the distro and/or what a user may have changed. I by the way have no evidence- just asking ;)

PR #837, opened 2019-08-23
feat(accounts_util): adding "add" flag to enable/disable the user creation process

afshinm:
## Overview
This PR adds a new boolean flag, `add: bool = True`, to google-accounts-daemon to enable/disable the user creation process (default set to True, so local users will be created if they don't exist)

## Purpose
Local users should not be created in some specific use-cases, for instance when LDAP is configured on the host. In this case, all users should be managed by LDAP and not the OS users database.

This flag is very similar to the `remove` flag which controls the derovisioning of users.

GCP support case number **19626907**

### DIFF
--- a/packages/python-google-compute-engine/google_compute_engine/accounts/accounts_daemon.py
+++ b/packages/python-google-compute-engine/google_compute_engine/accounts/accounts_daemon.py
@@ -39,7 +39,7 @@ class AccountsDaemon(object):
   user_ssh_keys = {}
 
   def __init__(
-      self, groups=None, remove=False, gpasswd_add_cmd=None,
+      self, groups=None, remove=False, add=True, gpasswd_add_cmd=None,
       gpasswd_remove_cmd=None, groupadd_cmd=None, useradd_cmd=None,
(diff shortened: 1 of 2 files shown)


illfelder:
Thanks for the contribution! We currently provide an option in the template config file to disable running the user accounts daemon. That might be more appropriate for users who are running their own LDAP infrastructure. Is there a reason that does not work for your needs?

afshinm:
We don't want to disable the accounts daemon since we need the `gcloud ssh/scp` feature as well. Basically, the user is _defined_ in our LDAP but we need accounts daemon to take care of the SSH key distribution.

afshinm:
ping @illfelder

illfelder:
Thanks for the additional information. We've been reviewing the proposal and considering the side effects associated with it. By disabling user creation, there is a lot of undefined behaviors associated with the accounts daemon and we don't feel comfortable making this change.

illfelder:
As a note, we do offer OS Login which provides LDAP-like user management while also managing SSH keys.

PR #846, opened 2019-09-10
Add missing DisableIpv6 function for SLES 12 and default case.

### LINKED ISSUE #844: e6eab0c1 breaks SUSE distros
With #e6eab0c1 the "DisableIpv6" interface was introduced in network_setup.py. This in turn call through to distribution specific implementations. However for SUSE the interface was not implemented and it was not added to the base class either. Therefore this generates a traceback:

google-networking: ERROR Exception calling the response handler. 'Utils' object has no attribute 'DisableIpv6'.
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/google_compute_engine/metadata_watcher.py", line 200, in WatchMetadata
    handler(response)
  File "/usr/lib/python2.7/site-packages/google_compute_engine/networking/network_daemon.py", line 97, in HandleNetworkInterfaces
    self.network_setup.DisableIpv6([default_interface.name])
  File "/usr/lib/python2.7/site-packages/google_compute_engine/networking/network_setup/network_setup.py", line 83, in DisableIpv6
    self.distro_utils.DisableIpv6(interfaces, self.logger)
AttributeError: 'Utils' object has no attribute 'DisableIpv6'

zmarano:
Fixes #844

### DIFF
--- a/packages/python-google-compute-engine/google_compute_engine/distro_lib/sles_12/utils.py
+++ b/packages/python-google-compute-engine/google_compute_engine/distro_lib/sles_12/utils.py
@@ -39,6 +39,15 @@ class Utils(utils.Utils):
     """
     pass
 
+  def DisableIpv6(self, interfaces, logger):
+    """Disable Ipv6.
+
+    Args:
+      interface: string, the output device names for enabling IPv6.
+      logger: logger object, used to write to SysLog and serial port.
+    """
+    pass
+
   def EnableNetworkInterfaces(self, interfaces, logger, dhclient_script=None):
     """Enable the list of network interfaces.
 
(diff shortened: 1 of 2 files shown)


rjschwei:
Thanks that was one way to about it. Getting in touch with the wicked team to see if we can follow the behavior implemented for dhclient

zmarano:
This will at least make it not fail. @hopkiw can talk about what our plans are to add functionality here or in the new guest agent.

rjschwei:
Received feedback from the wicked team. wicked does not support a feature to release a lease. Thus "pass" is the best we can do for now. I have created a feature request for wicked to gain this functionality.

PR #861, opened 2019-10-13
Wait on tentative v6 link-local addrs before DHCP

hopkiw:
If eth0 is in the middle of determining whether a self-assigned link-local IPv6 address is unique or not (duplicate address detection), dhclient IPv6 requests may fail, resulting in hosts without IPv6 addresses.

This process is supposed to take one second, but this commit adds a detection loop for this condition to be sure. Originally authored by Mark Hlady. dhclient bug: https://bugs.launchpad.net/ubuntu/+source/isc-dhcp/+bug/1633479

### DIFF
--- a/packages/python-google-compute-engine/google_compute_engine/distro_lib/helpers.py
+++ b/packages/python-google-compute-engine/google_compute_engine/distro_lib/helpers.py
@@ -17,6 +17,7 @@
 
 import os
 import subprocess
+import time
 
 
 def CallDhclient(
@@ -67,6 +68,24 @@ def CallDhclientIpv6(interfaces, logger, dhclient_script=None,
       logger.warning('Could not release IPv6 lease on interface %s.',
                      interfaces)
     return
+
+  # Check for a 'tentative' IPv6 address which would prevent `dhclient -6` from
+  # succeeding below. This should only take 1 second, but we try for up to 5.
+  command = [
+      'ip', '-6', '-o', 'a', 's', 'dev', 'eth0', 'scope', 'link', 'tentative'
+  ]
+  for i in range(5):
+    output = ""
+    try:
+      output = subprocess.check_output(command)
+    except subprocess.CalledProcessError as e:
+      logger.warning('Could not confirm tentative IPv6 address: %s.', e.output)
+    if output:
+      logger.info('Found tentative ipv6 link address %s, sleeping 1 second.',
+                  output.strip())
+      time.sleep(1)
+    else:
+      break
 
   if dhclient_script and os.path.exists(dhclient_script):
     dhclient_command += ['-sf', dhclient_script]
--- a/packages/python-google-compute-engine/google_compute_engine/distro_lib/helpers.py
+++ b/packages/python-google-compute-engine/google_compute_engine/distro_lib/helpers.py
@@ -71,11 +71,10 @@ def CallDhclientIpv6(interfaces, logger, dhclient_script=None,
 
   # Check for a 'tentative' IPv6 address which would prevent `dhclient -6` from
   # succeeding below. This should only take 1 second, but we try for up to 5.
-  command = [
-      'ip', '-6', '-o', 'a', 's', 'dev', 'eth0', 'scope', 'link', 'tentative'
-  ]
+  command = ['ip', '-6', '-o', 'a', 's', 'dev', interfaces[0], 'scope',
+             'link', 'tentative']
   for i in range(5):
-    output = ""
+    output = ''
     try:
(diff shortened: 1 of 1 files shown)


packages/python-google-compute-engine/google_compute_engine/distro_lib/helpers.py
```diff
@@ -68,6 +69,21 @@ def CallDhclientIpv6(interfaces, logger, dhclient_script=None,
                      interfaces)
     return
 
+  # Check for a 'tentative' IPv6 address which would prevent `dhclient -6` from
+  # succeeding below. This should only take 1 second, but we try for up to 5.
+  command = ['ip','-6','-o','a','s','dev','eth0','scope','link','tentative']
```
illfelder: Please add spaces after the commas in the list. Here's a copy pastable change:

```
  command = [
      'ip', '-6', '-o', 'a', 's', 'dev', 'eth0', 'scope', 'link', 'tentative']
```

packages/python-google-compute-engine/google_compute_engine/distro_lib/helpers.py
```diff
@@ -68,6 +69,21 @@ def CallDhclientIpv6(interfaces, logger, dhclient_script=None,
...
+      output = subprocess.check_output(command)
+    except subprocess.CalledProcessError as e:
+      logger.warning('Could not confirm tentative IPv6 address: %s.', e.output)
+    if output:
+      logger.info('Found tentative ipv6 link address %s, sleeping 1 second',
+                  output.strip())
+      time.sleep(1)
+    else:
+      break
+
```
illfelder: I think it would be worth moving this block into its own (tested) function. What do you think?

hopkiw: discussed offline and decided against it

packages/python-google-compute-engine/google_compute_engine/distro_lib/helpers.py
```diff
@@ -68,6 +69,23 @@ def CallDhclientIpv6(interfaces, logger, dhclient_script=None,
                      interfaces)
     return
 
+  # Check for a 'tentative' IPv6 address which would prevent `dhclient -6` from
+  # succeeding below. This should only take 1 second, but we try for up to 5.
+  command = ['ip', '-6', '-o', 'a', 's', 'dev'] + interfaces + ['scope', 'link',
```
illfelder: Please line wrap after the open bracket.

```
  command = ['ip', '-6', '-o', 'a', 's', 'dev'] + interfaces + [
      'scope', 'link', 'tentative']
```

packages/python-google-compute-engine/google_compute_engine/distro_lib/helpers.py
```diff
@@ -68,6 +69,23 @@ def CallDhclientIpv6(interfaces, logger, dhclient_script=None,
...
+                                                                'tentative']
+  for i in range(5):
+    output = ""
```
illfelder: Please use single quotes.

packages/python-google-compute-engine/google_compute_engine/distro_lib/helpers.py
illfelder: Will it work to run this command on multiple interfaces at the same time?

hopkiw: no, but the fact that this entire rigging uses the name 'interfaces' is misleading to begin with. these functions only pass around a list containing the default interface.
can add a for loop around the check here, if we want to extend the fiction that this feature supports multiple interfaces.

hopkiw: relevant code in NetworkDaemon.HandleNetworkInterfaces():
```
     if self.network_setup_enabled:                                              
       default_interface = network_interfaces[0]                                 
       if default_interface.ipv6:                                                
          self.network_setup.EnableIpv6([default_interface.name])                 
       else:                                                                     
         self.network_setup.DisableIpv6([default_interface.name])                
       self.network_setup.EnableNetworkInterfaces(                               
           [interface.name for interface in network_interfaces[1:]])  
```

illfelder: Interesting. I do thing that we should either ensure there is only one entry or loop over the list. Either is fine, but we shouldn't use the list in a way that would break us if the list ever contained more than one entry.

acornejo:
ping, could we get this submitted? i have seen more than 1 customer bit by this already

PR #866, opened 2019-11-12
Version:20191112.0

hopkiw:


zmarano:
Bump google-compute-engine as well (https://github.com/GoogleCloudPlatform/compute-image-packages/blob/development/packages/google-compute-engine/packaging/google-compute-engine.spec#L31)

hopkiw:
I think we can eliminate the historical version lock between these packages. For right now, explicitly declaring the current package dependency (and bumping packaging revisions, since no contents are changing).

Issue #869, opened 2019-11-15
GCP / CentOS "google_instance_setup" *always* regenerate SSH host key


arekkusu:
On Google Cloud running " /usr/bin/google_instance_setup" will *always* regenerate the SSH host key.

This concerns at least CentOS 6/7/8 after updating "python-google-compute-engine"

Problematic package: "python-google-compute-engine"  [1] version "20191112.00-g1.el7"
[python-google-compute-engine-20191112.00-g1.el7.noarch.rpm RPM DL](https://packages.cloud.google.com/yum/repos/google-compute-engine-el7-x86_64-stable/Packages/80d11aff4e3a486693371efa20b5a2e7c4c76a3668d1870979af795c1bb166ae-python-google-compute-engine-20191112.00-g1.el7.noarch.rpm)

Previous version:
[python-google-compute-engine-20190916.00-g1.el7.noarch.rpm RPM DL](https://packages.cloud.google.com/yum/repos/google-compute-engine-el7-x86_64-stable/Packages/f95dba61f66302af350789041fb30ddca7f2a3bb07ead4c517d73e8249b32130-python-google-compute-engine-20190916.00-g1.el7.noarch.rpm)

I believe this could be the problematic code as removing those line has an effect, but not sure.
https://github.com/GoogleCloudPlatform/compute-image-packages/commit/880daf9de49873d37aa6723dff4bba81542d1ecd#diff-478b5c2a4e03952901b1a18ea65b1b58L208-L211

[1] "python3-google-compute-engine" in the case of CentOS 8

arekkusu:
Reproduction steps in Google Cloud:
https://gist.github.com/arekkusu/f1472a712faf5099c21a322e0490b4cc

arekkusu:
Created PR https://github.com/GoogleCloudPlatform/compute-image-packages/pull/870

hopkiw:
This issue was resolved with a rollback. The feature was re-introduced without the typo in #874

PR #870, opened 2019-11-15
Should be "Instance" with uppercase "I"

arekkusu:
Fix for https://github.com/GoogleCloudPlatform/compute-image-packages/issues/869


hopkiw:
We performed a rollback of the relevant change, but this was the correct fix. Thanks!

Issue #862, opened 2019-10-16
system hangs on name resolution failure (startup or shutdown)


rjschwei:
Upon shutdown script_manager attemptes to retrieve data from the metadata server via the call chain
ScriptManager.RunScripts() -> retriever.GetScripts() -> watcher.GetMetadata()

While MetadataWatcher.GetMetadata() has arguments to avoid a forever loop in MetadataWatcher._HandleMetadataUpdate() by setting the "retry" argument to false this control option is not available in retriever.GetScripts(). Therefore it is not possible for the startup or shutdown sequence to succeed if name resolution fails.

Potential options for addressing the issue:
1.) Add the control options from MetadataWatcher.GetMetadata() to higher level implementations that call this code.
  + This would allow the ScriptManager.RunScripts() code to avoid an infinite loop as retriever.GetScripts(retry=False) could be used

2.) Add a retry counter to MetadataWatcher._HandleMetadataUpdate()
  + Have a set number of retries, maybe 50 to allow a potential transient network access issue to clear. This would introduce a delay in the case that showed this problem, but a system with persistent name resolution problems would still start or shutdown

3.) Explicitly ignore name resolution errors in MetadataWatcher._HandleMetadataUpdate()
  + This implies to abandon the wait loop when no name resolution is possible. However this also implies that if the issue is transient any data that may be important will not be considered.

IMHO 1 should be the preferred implementation approach but I didn't want to go down that path if there is not general consensus.

Please let me know and I'll be happy to provide a PR.

BongoEADGC6:
I've confirmed the newest version including fix #864 works with shutting down. Was able to build an image in Virtualbox via Packer without any intervention. This can be closed now.

hopkiw:
Thanks for the report @BongoEADGC6 

Also FYI: if you want to bypass shutdown scripts you can disable them with the instance_configs.cfg file or by `systemctl disable google-shutdown-scripts.service`, either one.

illfelder:
Reopening this since we are reverting the pull request.

As part of the instance initialization process, there is a non-deterministic amount of time before the instance will have access to the metadata server. Setting a retry limit may prevent startup scripts from executing, leaving the machine in an undefined (potentially broken) state. It is important that we do not have a retry limit for waiting for metadata during instance setup or before executing startup script. In a followup pull request, we can re-introduce the retry limits, setup functionality should not have limits by default.

PR #876, opened 2019-11-20
Roll forward a fix for #862.

illfelder:
Support retry limits in metadata retrieval.

### DIFF
--- a/packages/python-google-compute-engine/google_compute_engine/metadata_watcher.py
+++ b/packages/python-google-compute-engine/google_compute_engine/metadata_watcher.py
@@ -163,6 +163,7 @@ class MetadataWatcher(object):
       wait: bool, True if we should wait for a metadata change.
       timeout: int, timeout in seconds for returning metadata output.
       retry_limit: int or None, limit for number of times to retry on failure.
+          Retry indefinitely when set to None. Do not retry when set to zero.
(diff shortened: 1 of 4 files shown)


packages/python-google-compute-engine/google_compute_engine/metadata_watcher.py
```diff
@@ -202,18 +201,18 @@ def WatchMetadata(
...
-      self, metadata_key='', recursive=True, timeout=None, retry=True):
+      self, metadata_key='', recursive=True, timeout=None, retry_limit=None):
     """Retrieve the contents of metadata server for a metadata key.
 
     Args:
       metadata_key: string, the metadata key to watch for changes.
       recursive: bool, True if we should recursively watch for metadata changes.
       timeout: int, timeout in seconds for returning metadata output.
-      retry: bool, True if we should retry on failure.
+      retry_limit: int or None, limit for number of times to retry on failure.
```
wrigri: It could be useful to indicate here what happens when retry_limit is None (retry indefinitely ) and perhaps a note that setting retry_limit to 0 means "don't retry".

illfelder: Done.

PR #878, opened 2019-12-03
disk-expand: deb10 fixes

hopkiw:
* Include dmidecode (fixes confusing error message)
* Use correct function name for logging
* Don't bail on e2fsck errors; resize2fs includes an integrity pre check


hopkiw:
Do not merge, this needs some test improvements before we should ship it (given that the bugs fixed were not caught) but the bug fixes aren't colocated with the package.

PR #879, opened 2019-12-10
fix bugs in disk-expand

hopkiw:
* Include dmidecode (fixes confusing error message)
* Use correct function name for logging
* Don't bail on e2fsck errors; resize2fs includes an integrity pre check


packages/gce-disk-expand/packaging/gce-disk-expand.spec
```diff
@@ -23,6 +23,8 @@ Source0: %{name}_%{version}.orig.tar.gz
 Requires: e2fsprogs, dracut, grep, util-linux, parted
```
zmarano: Requires: e2fsprogs, dracut, grep, util-linux, parted, gdisk

packages/gce-disk-expand/packaging/gce-disk-expand.spec
```diff
@@ -23,6 +23,8 @@ Source0: %{name}_%{version}.orig.tar.gz
...
 Conflicts: dracut-modules-growroot
 
+BuildRequires: rsync
```
zmarano: We need rsync?

packages/gce-disk-expand/src/expandfs-lib.sh
```diff
@@ -60,32 +60,31 @@ blkid_get_fstype() (
...
-    parted -m ---pretend-input-tty "$disk" print Fix >/dev/null 2>&1 </dev/null
-    if parted -sm "$disk" print 2>&1 | grep -q "fix the GPT"; then
-      echo "Failed to fix the GPT."
-      return 1
-    fi
-  fi
+  local label=$(sgdisk_get_label "$disk")
+  [ "$label" != "gpt" ] && return
+
+  sgdisk --move-second-header "$disk"
```
zmarano: I believe you need to re-probe the partitions after this. At least that was what I found (partprobe).

packages/gce-disk-expand/src/usr/share/initramfs-tools/hooks/expand-rootfs
```diff
@@ -36,3 +36,4 @@ copy_exec /sbin/e2fsck /bin
 copy_exec /sbin/blkid /bin
 copy_exec /sbin/resize2fs /bin
 copy_exec /bin/udevadm /bin
+copy_exec /usr/sbin/dmidecode /bin
```
zmarano: What's dmidecode used for?

packages/gce-disk-expand/packaging/gce-disk-expand.spec
hopkiw: yes, it's used in this file

packages/gce-disk-expand/src/expandfs-lib.sh
hopkiw: what did you find? we aren't changing partitions here so i didn't find this to be the case

packages/gce-disk-expand/src/usr/share/initramfs-tools/hooks/expand-rootfs
hopkiw: parted i think calls it and complains without it

packages/gce-disk-expand/src/expandfs-lib.sh
zmarano: It was several weeks back but I believe that you would still get the error that the header was wrong unless the kernel was forced to re-read the partition table. I believe this was when I was trying to do a check with sgdisk however.

packages/gce-disk-expand/src/usr/share/initramfs-tools/hooks/expand-rootfs
zmarano: Oh yes I think you are right.

hopkiw:
working on:
debian-10 (gpt)
centos-7,rhel-7 (gpt, msdos)
centos-8,rhel-8 (gpt, msdos)

created 10g instance, installed package, resized disk to 30g, rebooted, confirmed 30g
this needs a test specifically to check that the condition where root is not the last partition does not get resized.

hopkiw:
blocking on GoogleCloudPlatform/oss-test-infra#330

PR #888, opened 2020-04-17
Fix disk expand for EL7 race conditions.

zmarano:
Proposed fix for disk expansion race condition.

### DIFF
--- a/packages/gce-disk-expand/src/usr/share/dracut/modules.d/50expand_rootfs/install
+++ b/packages/gce-disk-expand/src/usr/share/dracut/modules.d/50expand_rootfs/install
@@ -28,5 +28,4 @@ dracut_install grep
 dracut_install resize2fs
 dracut_install e2fsck
 dracut_install udevadm
-dracut_install systemctl
 dracut_install -o xfs_growfs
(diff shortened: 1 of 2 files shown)


packages/gce-disk-expand/src/usr/share/dracut/modules.d/50expand_rootfs/install
```diff
@@ -28,4 +28,5 @@ dracut_install grep
 dracut_install resize2fs
 dracut_install e2fsck
 dracut_install udevadm
+dracut_install systemctl
```
hopkiw, line 31: did it not work without this?? initrd should have its own systemctl, i'm unsure what the behavior is if we import the system systemctl binary

zmarano: I didn't try... this is technically what you should do for any of these things (its like declaring your package dependencies). In any case, it would not hurt as it would just re-copy the same thing if its done elsewhere.

zmarano: It appears from other modules that use it- that this is not required... I'll remove it.

PR #891, opened 2020-04-28
Interim fix fo EL7 GPT header move race condition with systemd-fsck.

zmarano:
This race manifests depending on how quickly the system boots and is more prevalent on some faster systems. When running the GPT move header operation multiple systemd-fsck tasks are spawned because of device enumeration. These can overrun each other and cause failures. Adding a sleep before and after this operation resolves the race. But we should find another solution that doesn't rely on sleep if possible.

Also add better console logging messages to the steps being run.

### DIFF
--- a/packages/gce-disk-expand/packaging/debian/changelog
+++ b/packages/gce-disk-expand/packaging/debian/changelog
@@ -1,3 +1,9 @@
+gce-disk-expand (1:20200428.00-g1) stable; urgency=medium
+
+  * Fix potential races in disk expand logic with GPT disks.
+
+ -- Google Cloud Team <gc-team@google.com>  Tue, 28 Apr 2020 10:24:07 -0700
+
 gce-disk-expand (1:20190708.00-g1) stable; urgency=medium
 
   * Update disk expand to fix bugs with GPT.
--- a/packages/gce-disk-expand/packaging/setup_deb.sh
+++ b/packages/gce-disk-expand/packaging/setup_deb.sh
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 NAME="gce-disk-expand"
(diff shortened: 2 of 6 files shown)


packages/gce-disk-expand/src/expandfs-lib.sh
```diff
@@ -78,7 +84,14 @@ sgdisk_fix_gpt() {
   local label=$(sgdisk_get_label "$disk")
   [ "$label" != "gpt" ] && return
 
+  # Add sleeps to allow this operation to fully complete. On some systems, other
+  # operations such as systemd-fsck tasks can collide and fail.
+  udevadm settle
+  sleep 1
```
illfelder, line 90: Please add a TODO to remove/replace these sleeps with better logic when we figure out a better alternative.

zmarano: Done

packages/gce-disk-expand/packaging/debian/changelog
```diff
@@ -1,3 +1,9 @@
+gce-disk-expand (1:20200428.00-g1) stable; urgency=medium
```
hopkiw, line 1: we don't edit these files anymore - they will be overwritten during build anyway

packages/gce-disk-expand/packaging/setup_deb.sh
```diff
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 NAME="gce-disk-expand"
-VERSION="20190708.00"
+VERSION="20200428.00"
```
hopkiw, line 17: the build system provides versions. this file and others like it are unused

PR #892, opened 2020-04-28
use flock in EL disk expand

hopkiw:


packages/gce-disk-expand/src/usr/share/dracut/modules.d/50expand_rootfs/expand_rootfs.sh
```diff
@@ -40,35 +40,39 @@ main() {
   disk=${out%:*}
   partnum=${out#*:}
 
-  if ! parted_needresize "$disk" "$partnum"; then
-    kmsg "Disk ${rootdev} doesn't need resizing"
-    return
-  fi
+  (
+    if ! flock -n 9; then
```
zmarano, line 44: Can you add a comment here describing why we are doing this (for future readers of this script at least).

zmarano: Also, do we need to install flock into dracut?

